PR #2, opened 2025-10-05
[SANC-14] Add all svg icons as memoed components

JustinTan-1:
Converted each SVG to memoed TSX components. 

https://codethechangeyyc.atlassian.net/browse/SANC-14

## Description
Added each SVG icon as a memoed TSX component. Each component can be resized with "height" and "width" prop.


burtonjong (approved):
lgtm

PR #3, opened 2025-10-05
SANC-11: configure fonts

burtonjong:
configured fonts.

### DIFF
--- a/src/styles/globals.scss
+++ b/src/styles/globals.scss
@@ -1,17 +1,48 @@
-html,
+@import "variables";
+
 body {
-	padding: 0;
-	margin: 0;
-	font-family:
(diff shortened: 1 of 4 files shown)


src/styles/globals.scss
```diff
@@ -1,17 +1,48 @@
...
+.heading-4 {
+  font-size: clamp(18px, 2vw, 16px);
+  font-weight: $font-weight-semibold;
+}
+
+p,
+.text-regular {
+  font-size: clamp(16px, 1.5vw, 14px);
+  font-weight: $font-weight-regular;
 }
```
burtonjong, line 48: this doesn't completely match what victoria gave in figma, but i did this on purpose. I'm going to talk to her about it the next time we meet.

this is just to get us started.

PR #5, opened 2025-10-05
SANC-13: add basic form

themaxboucher:
Create basic skeleton for the agency booking form. Inputted data gets logged to the console when submitted.

Screenshot 2025-10-05 at 4 48 43 PM


src/app/_components/agency-form.tsx
```diff
@@ -0,0 +1,109 @@
...
+            Logistics
+          </Title>
+
+          <TextInput
+            withAsterisk
+            label="Date and time of transport"
+            placeholder="YYYY-MM-DD HH:MM"
+            key={form.key("transportDateTime")}
+            {...form.getInputProps("transportDateTime")}
+          />
```
burtonjong: This should be a datetimepicker from mantine

burtonjong (changes requested):
Can you also merge main (update your branch with the latest changes)

src/app/_components/agency-form.tsx
```diff
@@ -0,0 +1,109 @@
...
+      transportDateTime: (value) => (value.trim().length > 0 ? null : "Date and time is required"),
+      purpose: (value) => (value.trim().length > 0 ? null : "Purpose is required"),
+      destinationAddress: (value) =>
+        value.trim().length > 0 ? null : "Destination address is required",
+    },
+  });
+
+  function handleSubmit(values: typeof form.values) {
+    console.log(values);
+  }
```
burtonjong: Can we create a global type here instead of `typeof form.values`, there's a chance that this form type will be reused elsewhere (like in our tRPC backend)

src/app/_components/agency-form.module.scss
burtonjong, line 1: I would say that we should probably just make these styles for a reusable datepicker, but that can be in another ticket. I made it here for reference whoever ends up taking this: https://codethechangeyyc.atlassian.net/browse/SANC-22

burtonjong (approved):
Other than my comment looks good

PR #7, opened 2025-10-19
SANC-9: Configure theme colors

tanvimahal:
added color variables and utility classes


promatty (approved):
lgtm, but do we want to use css vars for a darkmode? is that a feature we will want eventually @burtonjong ?

burtonjong:
We don't have to worry about it today

PR #11, opened 2025-10-31
SANC-35: add notification lib helper

burtonjong:
Did more than just the notifications here in this PR:

- added a configurable chevron with rotation props
- created skeleton for the agency page
- created notifications
     - error (red)
     - warning (yellow)
     - success (green)
     - info (blue)
     
     
Skeleton:

https://github.com/user-attachments/assets/5c802cba-0dfc-4e8a-88a5-985c09c4200a

Notifications:

image


src/app/page.tsx
```diff
@@ -71,6 +72,14 @@ export default async function Home() {
...
         </div>
 
+        <div>
+          <h1>Go to agency page</h1>
+          <Link href="/agency/home">Agency Home Page</Link>
+          <h1>Go to admin page (there is auth here)</h1>
+          <Link href="/admin/home">Admin Home Page</Link>
+          <h1>Go to drivers page (there is auth here)</h1>
+          <Link href="/driver/home">Drivers Home Page</Link>
+        </div>
```
burtonjong, line 85: temp buttons to navigate to pages

tsconfig.json
```diff
@@ -21,7 +21,7 @@
     "module": "ESNext",
     "moduleResolution": "Bundler",
     "jsx": "preserve",
-    "plugins": [{ "name": "next" }],
+    "plugins": [{ "name": "next" }, { "name": "typescript-plugin-css-modules" }],
```
burtonjong, line 24: i added this because we need typesense for scss. We will set this up on sunday

src/assets/icons/chevron.tsx
burtonjong, line 1: I believe we need a ticket as well for a button with only an icon? not sure if that's possible right now

PR #12, opened 2025-11-02
SANC-20: global mantine modal component

burtonjong:
image
I will delete the buttons eventually that are a part of the other form, was just not a part of this actual ticket.

demo:

https://github.com/user-attachments/assets/de5fb706-ba48-41db-8692-fff1e060212c

### DIFF
--- a/src/app/_components/common/modal/modal.tsx
+++ b/src/app/_components/common/modal/modal.tsx
@@ -1,0 +1,68 @@
+import { Button, Group, Modal as MantineModal } from "@mantine/core";
+import type { ReactNode } from "react";
+
+interface ModalProps {
+  opened: boolean;
+  onClose: () => void;
+  onConfirm: () => void;
+  title?: string;
+  children: ReactNode;
+  size?: string | number;
+  centered?: boolean;
+  withCloseButton?: boolean;
+  closeOnClickOutside?: boolean;
+  closeOnEscape?: boolean;
+  footer?: ReactNode;
+  showDefaultFooter?: boolean;
+  confirmText?: string;
+  cancelText?: string;
+  loading?: boolean;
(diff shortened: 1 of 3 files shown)


burtonjong:
Eventually when we press a button a notification should be fired as well in the bottom right based on if it was a success or not.

we will have to create a tanstack provider for this: https://tanstack.com/query/latest/docs/framework/vue/reference/useMutation

basically onsuccess or onfailure fire a toast.


Created a ticket for tanstack provider to fetch data: https://codethechangeyyc.atlassian.net/browse/SANC-37

src/app/_components/common/modal/modal.tsx
```diff
@@ -0,0 +1,68 @@
...
+  centered?: boolean;
+  withCloseButton?: boolean;
+  closeOnClickOutside?: boolean;
+  closeOnEscape?: boolean;
+  footer?: ReactNode;
+  showDefaultFooter?: boolean;
+  confirmText?: string;
+  cancelText?: string;
+  loading?: boolean;
+  overlayProps?: Record<string, string | number | boolean>;
```
promatty: overlayProps isnt used in your tests, but in case we ever do use it, the type you gave it is broader than what mantine expects.

import type { ModalBaseOverlayProps } from "@mantine/core";

interface ModalProps {
...
  overlayProps?: ModalBaseOverlayProps;
}

that's what mantine expects according to their docs
(https://mantine.dev/core/modal/?t=props)
Screenshot 2025-11-02 at 12 09 03 AM

promatty (changes requested):

src/app/_components/common/modal/modal.tsx
burtonjong: fixed

PR #13, opened 2025-11-04
SANC-38: add an icon button type that only has an icon with no text

wesleylui:
Here's what I did:

- add "icon" button variant
- ariaLabel is for accessibility/screen readers
- Development-time console warning remains for any icon-only button missing an accessible (ariaLabel) name.
- update style guide with icon button examples

image

### DIFF
--- a/src/app/_components/Button.module.scss
+++ b/src/app/_components/Button.module.scss
@@ -1,54 +1,54 @@
 @import "@/styles/variables";
 
 .primaryButton {
-	background-color: $color-primary;
-	color: $color-primary-text;
+  background-color: $color-primary;
+  color: $color-primary-text;
 
-	&:hover:not(:disabled) {
-		background-color: $color-primary-hover;
-		color: $color-primary-hover-text;
-	}
+  &:hover:not(:disabled) {
+    background-color: $color-primary-hover;
+    color: $color-primary-hover-text;
+  }
 }
 
 .secondaryButton {
-	background-color: $color-secondary;
-	color: $color-secondary-text;
-	border: 1px solid $color-light-grey;
+  background-color: $color-secondary;
+  color: $color-secondary-text;
+  border: 1px solid $color-light-grey;
 
-	&:hover:not(:disabled) {
-		background-color: $color-secondary-hover;
-		color: $color-secondary-text;
-	}
+  &:hover:not(:disabled) {
+    background-color: $color-secondary-hover;
+    color: $color-secondary-text;
+  }
 }
 
 .disabledButton {
-	background-color: $color-disabled;
-	color: $color-disabled-text;
-	border: none;
-	cursor: not-allowed;
+  background-color: $color-disabled;
+  color: $color-disabled-text;
+  border: none;
+  cursor: not-allowed;
 
-	&:hover {
-		background-color: $color-disabled;
-		color: $color-disabled-text;
-	}
+  &:hover {
+    background-color: $color-disabled;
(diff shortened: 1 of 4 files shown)


src/app/_components/Button.tsx
```diff
@@ -28,22 +29,35 @@ export default function Button({
...
   const isPrimary = variant === "primary";
+  const isIconOnly = variant === "icon" || (!text && !!icon);
+
+  // Accessibility enforcement (development only)
+  if (isIconOnly && !ariaLabel && process.env.NODE_ENV !== "production") {
+    // eslint-disable-next-line no-console
+    console.warn(
+      "Button: icon-only button rendered without `ariaLabel` — provide `ariaLabel` for screen reader accessibility.",
+    );
+  }
```
burtonjong: if you make `ariaLabel` a required prop then you don't need to do this and they'll get an error and won't be able to push their changes

src/app/_components/Button.tsx
```diff
@@ -28,22 +29,35 @@ export default function Button({
   icon,
   onClick,
   disabled = false,
+  ariaLabel,
 }: ButtonProps) {
   const isPrimary = variant === "primary";
+  const isIconOnly = variant === "icon" || (!text && !!icon);
```
burtonjong: to determine if primary secondary/ icon only you can just create a helper function, sometihng like `getButtonStyles` and then call that to pass into classNames instead of using all these ternaries like on L56

burtonjong (changes requested):
image


I like having the buttons that are icon only, but having anohte roption to have no background at all (transparent) would be nice as well. That is what we are using here to either get the day/week view.

I'm going to leave this up to you, Mantine has an action button component: https://mantine.dev/core/action-icon/

Maybe we can use this instead of doing our own thing and adding an icon like this.

src/app/_components/Button.module.scss
```diff
@@ -1,34 +1,54 @@
 @import "@/styles/variables";
```
promatty, line 1: please change indentation back to 2 spaces

promatty (dismissed):

wesleylui:
what the iconButtons look like now
image

one problem:
- a button cannot be disabled AND transparent
because of the ordering in scss (disabled comes before transparent)
transparency can be applied on any type of button :D

burtonjong:
> what the iconButtons look like now image
> 
> one problem:
> 
> * a button cannot be disabled AND transparent
>   because of the ordering in scss (disabled comes before transparent)
>   transparency can be applied on any type of button :D

I'm pretty sure we are able to make the background disabled looking background dissapear when disabled and transparent using !important but i think thats fine. It's not that big of a deal

src/app/_components/IconButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const getIconButtonClass = () => {
+    const classes = [styles.iconButton];
+    if (disabled) {
+      classes.push(styles.disabledButton);
+    }
+    if (transparent) {
+      classes.push(styles.transparentButton);
+    }
+    return classes.join(" ");
+  };
```
burtonjong: in `Button.tsx` you define the helper function outside of the actual Button component. Let's keep it consistent and do the same and move this out

burtonjong (changes requested):

burtonjong (approved):

burtonjong:
Feel free to merge @wesleylui

PR #15, opened 2025-11-05
Sanc-26 implement trip creation api endpoints 

JustinTan-1:
Created API endpoint for creating trips on the DB. User must be authenticated as an "admin" or as an "agency" before being allowed to submit. 

# Testing
If User has the correct permissions:
DatabaseSnippet

If user is not authenticaled API will not execute
image


src/server/api/trpc.ts
```diff
@@ -130,3 +130,17 @@ export const protectedProcedure = t.procedure.use(timingMiddleware).use(({ ctx,
     },
   });
 });
+
+export const adminProcedure = t.procedure.use(timingMiddleware).use(({ ctx, next }) => {
```
burtonjong, line 135: looks good eventually in prod we will remove this timing middleware though

src/server/api/trpc.ts
```diff
@@ -130,3 +130,17 @@ export const protectedProcedure = t.procedure.use(timingMiddleware).use(({ ctx,
...
+  const userRole = ctx.session?.user.role;
+  const allowedRoles = ["agency", "admin"];
```
burtonjong: can we make it like this instead:

export const ADMIN_ROLES = [UserRole.AGENCY, UserRole.ADMIN] as const;

and then use ADMIN_PROCEDURE_ROLES instead of allowedRoles?

You can define this in types.ts

burtonjong (changes requested):
looks good just one change

burtonjong (approved):
LG

burtonjong (approved):
LG

PR #16, opened 2025-11-06
SANC-21: skeleton for top nav

tanvimahal:
created a common navbar for 3 different views (admin, agency and driver). can pass an agency name to display at the top on the agency page.

also added a user icon

Screenshot 2025-11-06 165359
Screenshot 2025-11-06 165446
Screenshot 2025-11-06 165509

### DIFF
--- a/src/app/_components/navbar.tsx
+++ b/src/app/_components/navbar.tsx
@@ -1,0 +1,53 @@
+"use client";
+
+import Bell from "@/assets/icons/bell";
+import Home from "@/assets/icons/home";
+import User from "@/assets/icons/user";
+import { Group, Text } from "@mantine/core";
+import Button from "./Button";
+
+type NavbarView = "admin" | "agency" | "driver";
+
+interface NavbarProps {
+  view: NavbarView;
+  agencyName?: string; //only used in Agency View
+}
+
+export default function Navbar({ view, agencyName }: NavbarProps) {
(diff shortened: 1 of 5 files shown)


src/app/_components/navbar.tsx
```diff
@@ -0,0 +1,53 @@
+"use client";
+
+import Bell from "@/assets/icons/bell";
+import Home from "@/assets/icons/home";
+import User from "@/assets/icons/user";
+import { Group, Text } from "@mantine/core";
+import Button from "./Button";
```
promatty, line 7: please rebase first so you actually have the button component 😀

promatty (changes requested):

src/app/_components/navbar.tsx
tanvimahal: sorry it should be good now

burtonjong (changes requested):
Image

If i add the navbar that you created to the /agency/home page, here is what it looks like above

1 .I think that we want to remove the left and right padding of the header component because the overal padding of the page accounts for it:

Image

2. In the figma mockup the rider logs, vehicle logs, and schedule are not surrounded by a border. Lets remove that border

3. There is a line separating the header and the main content. You can add this just by making border bottom 1px sollid black

burtonjong (approved):

PR #18, opened 2025-11-14
SANC-39: Google maps link integration

jason-duong4509:
- Added front-end and back-end verification checks for when the agency adds a destination address to a form
- Integrated Google maps into a button, such that when the button represents a location, it can be clicked to bring the user to Google maps

Images:
image
- Backend rejects an input as the address is too vague

image

- Google maps suggested locations based on user input

image

- Frontend rejects an input because the user did not select a Google-suggested location

image

- What the user sees upon clicking a button with a location in it on desktop 

## Summary by CodeRabbit

* **New Features**
  * Google Maps Places Autocomplete integration for address input fields with location suggestions.
  * Real-time address validation to ensure accurate destination addresses.
  * Added "View in Google Maps" link to view locations directly on the map.


src/app/_components/agencycomponents/agency-interactive-area.tsx
```diff
@@ -19,6 +19,31 @@ export const AgencyInteractiveArea = ({ initialViewMode = ViewMode.CALENDAR }: P
...
+  //Define a variable that react will reassign its value on runtime
+  //Starts off as null but will equal (through inputElement.current) an HTML input element when assigned at runtime
+  //Will be assigned an HTML input element when the mantine form loads
+  const inputElement = useRef<HTMLInputElement | null>(null);
+
+  //Run the following once, after react renders everything
+  useEffect(() => {
+    //Create an HTML script element so that the google maps API is loaded at runtime (needed for billing reasons among others)
+    const googleScript = document.createElement("script");
+    googleScript.src = `https://maps.googleapis.com/maps/api/js?key=${process.env.NEXT_PUBLIC_GOOGLE_MAPS_API_KEY}&libraries=places&v=beta&loading=async`;
```
promatty: we must ensure the prod key is set up with api restrictions in google cloud console. 
ie. Application restrictions:
   → HTTP referrers (websites)
   → Add your domain: https://yourdomain.com/*
   → Add localhost for dev: http://localhost:3000/*

  API restrictions:
   → Restrict key
   → Select ONLY:
      • Maps JavaScript API
      • Places API (optional, for autocomplete)

src/server/api/routers/form.ts
```diff
@@ -107,4 +107,36 @@ export const formRouter = createTRPCRouter({
...
+
+  //TODO: Change from publicProcedure to protectedProcedure (it is the former for testing until authentication is made)
+  validateDestinationAddress: publicProcedure
+    .input(z.object({ regionCode: z.string(), destinationAddress: z.array(z.string()) }))
+    .mutation(async ({ input }) => {
+      const { regionCode, destinationAddress } = input; //Grab passed variables
+
+      //Make an API call to Google Maps API to validate the inputs
+      const response = await fetch(
+        `https://addressvalidation.googleapis.com/v1:validateAddress?key=${process.env.NEXT_PUBLIC_GOOGLE_MAPS_API_KEY}`,
```
promatty: the backend api call should be using a seperate, PRIVATE key!! users can see NEXT_PUBLIC api keys in their network tabs (and thus run our pockets)

src/server/api/routers/form.ts
```diff
@@ -107,4 +107,36 @@ export const formRouter = createTRPCRouter({
...
   }),
+
+  //TODO: Change from publicProcedure to protectedProcedure (it is the former for testing until authentication is made)
+  validateDestinationAddress: publicProcedure
+    .input(z.object({ regionCode: z.string(), destinationAddress: z.array(z.string()) }))
+    .mutation(async ({ input }) => {
+      const { regionCode, destinationAddress } = input; //Grab passed variables
+
+      //Make an API call to Google Maps API to validate the inputs
+      const response = await fetch(
```
promatty: we need error handling incase this fetch fails

src/server/api/routers/form.ts
```diff
@@ -107,4 +107,36 @@ export const formRouter = createTRPCRouter({
...
+            },
+          }),
+        },
+      );
+
+      //Turn the response object into a JSON to check the result
+      const data = await response.json();
+
+      //If the address is good (addressComplete === true) then return Google's formatted version of the address, else false
+      return data.result.verdict.addressComplete === true
```
promatty: need checks to ensure the response from google api is even okay. 
```suggestion
      if (!data?.result?.verdict) {
        throw new Error('Invalid response from Google API');
  }
```

src/server/api/routers/form.ts
```diff
@@ -107,4 +107,36 @@ export const formRouter = createTRPCRouter({
...
+        ? data.result.address.formattedAddress
+        : false;
```
promatty: im not too sure about this, but maybe instead of false it should be null

src/app/_components/agencycomponents/agency-interactive-area.tsx
```diff
@@ -37,13 +62,70 @@ export const AgencyInteractiveArea = ({ initialViewMode = ViewMode.CALENDAR }: P
...
+      setValidationAddressGood(false); //User started typing, assume input is invalid
+    }
+    inputElement.current.addEventListener("input", inputElementOnInput);
+
+    //When inputElement.current changes, remove the listeners from old elements to prevent any performance issues
+    return () => {
+      google.maps.event.clearInstanceListeners(googleAutoCompleteElement); //Removes the place_changed event listener to the old element
+      inputElement.current?.removeEventListener("input", inputElementOnInput); //Removes the listener defined as inputElementOnInput from inputElement.current before replacing it
+    };
+  }, [inputElement.current]);
```
promatty, line 112: react refs dont trigger re-render. perhaps we should use googleScriptLoaded

src/app/_components/agencycomponents/agency-interactive-area.tsx
```diff
@@ -37,13 +62,70 @@ export const AgencyInteractiveArea = ({ initialViewMode = ViewMode.CALENDAR }: P
       transportDateTime: (value) => (value.trim().length > 0 ? null : "Date and time is required"),
       purpose: (value) => (value.trim().length > 0 ? null : "Purpose is required"),
       destinationAddress: (value) =>
-        value.trim().length > 0 ? null : "Destination address is required",
+        value.trim().length > 0
+          ? validationAddressGood
+            ? null
```
promatty, line 61: @burtonjong the current validation means the user MUST select from the google autocomplete dropdown. do we think there will ever be an address which isnt on google maps? ie. what if they want to input a specific unit of a building?

package-lock.json
```diff
@@ -0,0 +1,5411 @@
+{
```
promatty: we are using yarn, not npm

promatty (changes requested):

src/app/_components/agencycomponents/agency-interactive-area.tsx
burtonjong: I think general location is enough. If they complain we fix it later.

src/server/api/routers/form.ts
```diff
@@ -107,4 +107,53 @@ export const formRouter = createTRPCRouter({
...
+  //TODO: Change from publicProcedure to protectedProcedure (it is the former for testing until authentication is made)
+  validateDestinationAddress: publicProcedure
+    .input(z.object({ regionCode: z.string(), destinationAddress: z.array(z.string()) }))
+    .mutation(async ({ input }) => {
+      const { regionCode, destinationAddress } = input; //Grab passed variables
+
+      try {
+        //Make an API call to Google Maps API to validate the inputs
+        const response = await fetch(
+          `https://addressvalidation.googleapis.com/v1:validateAddress?key=${process.env.GOOGLE_MAPS_API_KEY}`,
```
burtonjong: You'll have to actually add these to `env.js`. Then you can import `src/@env` or whatever it is and directly reference env.GOOGLE_MAPS_API_KEY

src/server/api/routers/form.ts
```diff
@@ -107,4 +107,53 @@ export const formRouter = createTRPCRouter({
...
+
+        //Check google's sent data is valid before returning
+        if (!data?.result?.verdict) {
+          throw new Error("Invalid response from Google API");
+        }
+
+        //If the address is good (addressComplete === true) then return Google's formatted version of the address, else null
+        return data.result.verdict.addressComplete === true
+          ? data.result.address.formattedAddress
+          : null;
```
burtonjong, line 146: I mean does this even need to be a conditional? Do you get a type error if you don't have the ternary?

src/app/_components/agencycomponents/agency-interactive-area.tsx
```diff
@@ -19,6 +19,31 @@ export const AgencyInteractiveArea = ({ initialViewMode = ViewMode.CALENDAR }: P
...
+  const [googleScriptLoaded, setInitialLoad] = useState<boolean>(false); //Initially, the script has not loaded
+  const [validationAddressGood, setValidationAddressGood] = useState<boolean>(false);
+
+  //Define a variable that react will reassign its value on runtime
+  //Starts off as null but will equal (through inputElement.current) an HTML input element when assigned at runtime
+  //Will be assigned an HTML input element when the mantine form loads
+  const inputElement = useRef<HTMLInputElement | null>(null);
+
+  //Run the following once, after react renders everything
+  useEffect(() => {
```
burtonjong: I don't like having to use a useeffect here. Take a look at this article: https://stackoverflow.com/questions/72571603/how-to-load-google-maps-api-using-nextjs

next js also has a <Script/> component: https://nextjs.org/docs/app/api-reference/components/script

try and find an alternative ...

burtonjong (changes requested):
some comments

src/server/api/routers/form.ts
burtonjong: Just kind of tell me the flow here of different usecases that can happen im just confused

src/app/_components/agencycomponents/agency-interactive-area.tsx
jason-duong4509: Resolved: API keys used for development are tested for functionality with these restrictions set

src/server/api/routers/form.ts
jason-duong4509: Resolved: The backend uses its own private API key

src/server/api/routers/form.ts
jason-duong4509: Resolved: A try catch was used to handle errors from the fetch function

src/server/api/routers/form.ts
jason-duong4509: Resolved: Added the code suggestion in a try catch

src/server/api/routers/form.ts
jason-duong4509: Resolved: Changed the return to null

package-lock.json
jason-duong4509: Resolved: Got rid of NPM and used yarn

src/server/api/routers/form.ts
jason-duong4509: I'm not sure what you mean by mean with the type error arising from not having a ternary.

The flow of logic is:
1. The backend sends a fetch to Google's address validation API
2. It gets a bunch of stuff in JSON format
3. data.result.verdict.addressComplete is always true if the API was able to confirm that the given input is very well formatted (Which it should be if user input is autocorrected by the frontend)
   - If this is true, we return data.result.address.formattedAddress. This is just Google's version of the user's input. It is helpful to use their format to store in the DB as it's well formatted. It also helps in other features that may want to work with this address (like getting a placeID based on the address)
4. If data.result.verdict.addressComplete is not true (Google does not think it's a well formatted address), return null 

I believe it should be a conditional as we need to determine whether the address is well formatted, based on the API call

I used these for reference: 
- https://developers.google.com/maps/documentation/address-validation/understand-response
- https://developers.google.com/maps/documentation/address-validation/reference/rest/v1/TopLevel/validateAddress#address

src/app/_components/agencycomponents/agency-interactive-area.tsx
jason-duong4509: Resolved: I removed the useEffect() when loading the Google Maps API and opted to use their useLoadScript() (from your stack overflow URL)

src/server/api/routers/form.ts
jason-duong4509: Resolved: I changed all instances where a Google Maps API key was used. They now grab their keys from env.js and not through process.env

src/app/_components/agencycomponents/agency-interactive-area.tsx
jason-duong4509: I should mention that the private key requires different restrictions:
- It cannot have the HTTP referrers restriction but should either have none or by IP
- For this ticket, the private key is only used for address validation, so it only needs the Address Validation API. Places API and Maps JavaScript API are not required for the private key

src/app/_components/agencycomponents/agency-interactive-area.tsx
```diff
@@ -68,6 +161,11 @@ export const BookingInteractiveArea = ({ initialViewMode = ViewMode.CALENDAR }:
     }, 2000);
   };
 
+  //If the script hasn't loaded yet, don't render anything until it does
+  if (!isLoaded) {
+    return <></>;
+  }
```
burtonjong, line 167: Can't we just return a "Loading..."  or something instead of nothing?

burtonjong (approved):

promatty (approved):
lets just get this in. remember to address burton's comment in your next pr

PR #17, opened 2025-11-11
SANC-16: add calendar view

themaxboucher:
Create the calendar view for the agency page. 

Screenshot 2025-11-11 at 12 08 39 PM

## Summary by CodeRabbit

* **New Features**
  * Interactive calendar with day/week modes and responsive behavior
  * Date navigation controls to browse days/weeks
  * Bookings show status indicators (completed, in progress, incomplete) and open detailed info on click
* **Style**
  * Updated calendar styling and typography for clearer layout and event presentation
* **Chores**
  * Introduced booking/status data structures to support calendar features

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #20, opened 2025-11-17
SANC-41: Add scheduled date time field to booking schema

JustinTan-1:
Requires a ISO 8601 datetime string including Timezone.

## Summary by CodeRabbit

* **New Features**
  * Trip bookings now support start and end time specifications. Users can define precise scheduling windows with full timezone awareness when creating trips. Time information is automatically recorded with each booking.

### DIFF
--- a/src/server/db/booking-schema.ts
+++ b/src/server/db/booking-schema.ts
@@ -16,7 +16,11 @@ export const bookings = pgTable("bookings", {
   agencyId: text("agency_id")
     .notNull()
     .references(() => user.id, { onDelete: "cascade" }),
-  scheduledTime: timestamp("scheduled_time", {
+  startTime: timestamp("start_time", {
+    mode: "string",
+    withTimezone: true,
+  }).notNull(),
+  endTime: timestamp("end_time", {
     mode: "string",
     withTimezone: true,
   }).notNull(),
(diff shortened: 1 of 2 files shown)


src/server/db/booking-schema.ts
```diff
@@ -8,15 +8,18 @@ export const bookings = pgTable("bookings", {
   pickupLocation: text("pickup_location").notNull(),
   dropoffLocation: text("dropoff_location").notNull(),
   purpose: text("purpose"),
-  passengerInfo: text("passengerInfo").notNull(),
+  passengerInfo: text("passenger_info").notNull(),
```
burtonjong, line 11: why this change here?

src/server/db/booking-schema.ts
burtonjong, line 1: This should match what max has here in the `types.ts` file:

https://github.com/Code-the-Change-YYC/salvationarmy/pull/17/files#diff-11e9171ff36dfa6e6b396909ac6b924cf79337d5f05a7eaa15f78a9b2e2eded6R54-R57

there is a start and end, not scheduled time. Maybe figure out what would be easiest - computing the start and time and inserting that, or only having " scheduled time". Make sure other fields match what they are supposed to as well!

burtonjong (changes requested):

src/server/db/booking-schema.ts
JustinTan-1: It was a minor styling change to make it consistent with the other fields

src/server/db/booking-schema.ts
JustinTan-1: I added string fields for start and end times, let me know if there is anything else to change

burtonjong (approved):

PR #24, opened 2025-11-24
Match booking form values to API

JustinTan-1:
- Added title field and pickupAddress to booking form and matched values to API

TODO: Still need Start Date and End date passed to API from google maps calculations

## Summary by CodeRabbit

* **New Features**
  * Added Booking Name field and separate Start Time / Arrival Time inputs on the agency booking form.
  * Added Pickup Address input and explicit resident/contact fields for passenger details.

* **Bug Fixes**
  * Strengthened validation for title, times, pickup/destination addresses, and resident/contact info.

* **Style**
  * Reorganized booking form layout and updated calendar/event displays to show Pickup and Dropoff addresses.


src/server/api/routers/trip.ts
```diff
@@ -8,21 +8,23 @@ export const tripRouter = createTRPCRouter({
...
+        destinationAddress: z.string(),
       }),
     )
     .mutation(async ({ ctx, input }) => {
       await ctx.db.insert(bookings).values({
         title: input.title,
-        pickupLocation: input.pickupLocation,
-        dropoffLocation: input.dropoffLocation,
-        passengerInfo: input.passengerInfo,
+        pickupLocation: input.pickupAddress,
```
burtonjong: I feel like the names of these should match... @JustinTan-1 can you make it match whatever's in the figma? for all of these fields here?

src/app/_components/agencycomponents/agency-interactive-area.tsx
```diff
@@ -42,19 +42,23 @@ export const BookingInteractiveArea = ({ initialViewMode = ViewMode.CALENDAR }:
   const form = useForm({
     mode: "uncontrolled",
     initialValues: {
+      title: "",
       residentName: "",
       contactInfo: "",
       additionalInfo: "",
       transportDateTime: "",
```
burtonjong: just a note here @Lujarios , when you calculate the distance, please split it in a start and end time when you call your function to create a booking

burtonjong (changes requested):

src/app/_components/agencycomponents/agency-interactive-area.tsx
promatty: the backend expects both start and end time anyway. lets turn transportDateTime into start and end here.

src/server/api/routers/trip.ts
```diff
@@ -8,21 +8,23 @@ export const tripRouter = createTRPCRouter({
...
+        dropoffLocation: input.destinationAddress,
+        passengerInfo: `${input.residentName} ${input.contactInfo} ${input.additionalInfo}`,
```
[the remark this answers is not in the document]
promatty: this is a good point

promatty (changes requested):

burtonjong:
@JustinTan-1 I would read over some of code rabbit's comments. Especially the one about how endTime and startTime are Date objects

src/server/api/routers/trip.ts
JustinTan-1: I got the fields to match, but they are not the exact same as the Figma so lmk if that is a problem

JustinTan-1:
image
image

From what I am seeing, I think the AI might be tripping. The form collects string values for start and end time. Correct me if I am wrong though.

burtonjong:
> image image
> From what I am seeing, I think the AI might be tripping. The form collects string values for start and end time. Correct me if I am wrong though.

image


I console.logged the output... It is true that it's a string, but `"2025-11-27 00:00:00"` can't easily be turned into a string

I would say this is fine for now.

burtonjong (approved):

burtonjong:
@Yemyam As a note since you're doing https://codethechangeyyc.atlassian.net/browse/SANC-22, please also fulfill these requirements:

1. I would like to have datetimepicker so that we are able to actually choose the time by clicking, not just typing like how it currently is
https://mantine.dev/dates/date-time-picker/#timepicker-props

2. Whenever we submit the form, whatever value that is returned from the datetime picker should be a valid ISO string so that we can easily store it to what supabase expects which I think is an ISo string? I'm not 100% and that's something you can check

promatty:
if our time output looks like "2025-11-27 00:00:00", but the api expects an iso 8601 (ie "2025-11-27T00:00:00Z") wont this be an issue?

burtonjong:
> if our time output looks like "2025-11-27 00:00:00", but the api expects an iso 8601 (ie "2025-11-27T00:00:00Z") wont this be an issue?

Yes, which will be fixed once ethan is done his ticket

promatty (approved):

PR #25, opened 2025-11-25
SANC-49: When the user logs in redirect them to their respective homepage

tanvimahal:
Uses the session to confirm whether the user's logged-in, then retrieves their role from the session to route them to their home page based on their role.

Questions:

- Should I be using useActiveOrganization() and redirecting to the role/slug/home instead? 
for example: should it redirect to agency/agency-two/home or admin/admins/home
Or is it fine redirecting to the role (admin, agency, driver) homepage and just having the navbar show different agency names?

- When a user who is already logged in visits the /login page, there is a delay (1-2 seconds) where the login form shows up before redirecting to their homepage. Do you think there should be a loading,redirecting,etc. box that shows up instead of the login form?

## Summary by CodeRabbit

* **New Features**
  * Notification-based sign-in feedback replaces alerts for success and error.
  * Centralized, role-based dashboard redirect after login sends users to their appropriate home page.

* **Bug Fixes**
  * Driver users can now access the Driver home page (role-based access updated).


burtonjong:
@tanvimahal those are both really good questions

For 1. yes we should be redirecting to their slug. You can do this in next.js with dynamic routes: https://nextjs.org/docs/pages/building-your-application/routing/dynamic-routes

I would say that for admins/drivers, they can just be directed to the regular home page of `admin/home` and `driver/home` because they share it

However, Victoria mentioned agencies should only be able to view the details of their own booking and not others. You'll be able to see the booking itself (but not the name of passeenger, location, etc).

For agencies, it should redirect to 'agency/home/[slug]`

For 2. @promatty can you make a ticket to create a generic loading screen that can be reused? Tanvi, you can just leave this for now and this can be a different ticket :)

I would also say listen to some of code rabbits suggestions

src/app/login/page.tsx
```diff
@@ -3,14 +3,18 @@
...
 import Button from "@/app/_components/common/button/Button";
 import { authClient } from "@/lib/auth-client";
+import type { AuthUser } from "@/types/types";
 
 export default function LoginPage() {
   const [loading, setLoading] = useState(false);
   const router = useRouter();
 
+  const { data: session, isPending: sessionLoading } = authClient.useSession();
+  // const { data: activeOrganization, isPending: orgLoading } = authClient.useActiveOrganization();
```
burtonjong: And yes, useActiveOrganization() should be used over useSession()

burtonjong (changes requested):

src/app/login/page.tsx
```diff
@@ -26,19 +39,48 @@ export default function LoginPage() {
...
 
     try {
-      const { error } = await authClient.signIn.email({
+      const { data, error } = await authClient.signIn.email({
         email: values.email,
         password: values.password,
       });
 
       if (error) {
         alert(error.message || "Failed to sign in");
```
burtonjong: You can use the `notify` component that i made to instead show that the login failed, instead of showing a JS alert like this

src/app/login/page.tsx
```diff
@@ -26,19 +39,48 @@ export default function LoginPage() {
...
+        case Role.DRIVER:
+          router.push("/driver/home");
+          break;
+        default:
+          router.push("/");
+          break;
+      }
+    },
+    [router],
+  );
```
burtonjong: I don't mind the logic here, but I think we should be using better auth after hooks here: https://www.better-auth.com/docs/concepts/hooks#ctx

Here's kind of what I asked claude to generate:
This is probably a good starting point that you can clean up
```
export const auth = betterAuth({
  // ... your other config
  hooks: {
    after: createAuthMiddleware(async (ctx) => {
      // Only handle sign-in endpoints
      if (ctx.path.startsWith("/sign-in")) {
        const newSession = ctx.context.newSession;
        
        if (newSession?.user) {
          const role = newSession.user.role as Role;
          const activeOrgSlug = newSession.user.activeOrganization?.slug; // Adjust based on your session structure
          
          // Determine redirect path based on role
          let redirectPath = "/";
          
          switch (role) {
            case Role.ADMIN:
              redirectPath = "/admin/home";
              break;
            case Role.AGENCY:
              redirectPath = `/agency/home/${activeOrgSlug}`;
              break;
            case Role.DRIVER:
              redirectPath = "/driver/home";
              break;
          }
          
          // Redirect after successful sign-in
          throw ctx.redirect(redirectPath);
        }
      }
    }),
  },
});
```

This is in `auth.ts` btw

burtonjong: Basically if the sign in is successful it will call this hook. If it's not, then it will notify the user of an error

src/types/types.ts
```diff
@@ -95,3 +95,14 @@ export interface CalendarEvent {
     updatedBy?: string;
   };
 }
+
+export interface AuthUser {
```
burtonjong: Don't think this is needed, you can infer a type directly from the schema if you need: https://orm.drizzle.team/docs/goodies

using `InferSelectModel`

burtonjong (changes requested):

src/app/login/page.tsx
```diff
@@ -26,6 +38,7 @@ export default function LoginPage() {
   // todo2: we need to evaluate what org they are a part of to redirect them properly
   // todo3: after we handle the above, we need to make it so that this redirects them IF they
```
promatty: todo3 is still an issue; a logged in user doesnt get redirected to their homepage, they can still see the login page.

src/server/api/routers/organizations.ts
```diff
@@ -8,10 +8,67 @@ import {
...
+
+        if (userSession?.activeOrganizationId) {
+          const organization = await ctx.db.query.organization.findFirst({
+            where: (org, { eq }) => eq(org.id, userSession.activeOrganizationId!),
+          });
+          if (organization?.slug) {
+            redirectUrl = `/agency/home/${organization.slug}`;
+          }
+        }
+        break;
```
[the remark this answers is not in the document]
promatty: yeah this is an issue, we want to show the user an error message that org couldnt be found, or that there isnt a slug. maybe throw a trpcerror, not just a console log which coderabbit suggested

promatty (changes requested):
also delete the todo comments when youre done

src/server/api/routers/organizations.ts
```diff
@@ -8,10 +8,85 @@ import {
...
+    const role = user.role as Role;
+    let redirectUrl = "/";
+
+    switch (role) {
+      case Role.ADMIN:
+        redirectUrl = "/admin/home";
+        break;
+
+      case Role.AGENCY: {
+        const userSession = await ctx.db.query.session.findFirst({
```
promatty: using the most recent session from the db might actually cause issues if the user logs in on multiple devices. for example, 
10:00 AM - User logs in on laptop → Session A created
10:05 AM - User logs in on phone → Session B created (most recent)
10:10 AM - User makes request from laptop
What happens:
-Request from laptop sends Session A token (in cookie)
-Your query returns Session B (most recently created), and the user will be redirected based on session B's org. this is quite an edge case, but we should fix it regardless

```suggestion
case Role.AGENCY: {
  // use the session from the CURRENT request
  const activeOrgId = ctx.session.activeOrganizationId;
  
  if (!activeOrgId) {
    throw new TRPCError({
      code: "NOT_FOUND",
      message: "No active organization found for this agency user",
    });
  }
  
  // ... rest of logic
}
```

src/app/login/page.tsx
```diff
@@ -3,15 +3,34 @@
...
+    onError: (error) => {
+      notify.error(error.message || "Failed to get redirect URL");
+    },
+  });
+
+  useEffect(() => {
+    if (!sessionLoading && session) {
+      mutate();
+    }
+  }, [session, sessionLoading, mutate]);
```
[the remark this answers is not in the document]
promatty: yeah and fix this using isPending too

promatty (changes requested):
looks good after these

promatty:
oh and react/respond to burton's comments from last month if theyre resolved

src/app/login/page.tsx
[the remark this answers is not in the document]
tanvimahal: @promatty I already fixed the duplicate/multiple notifications before by removing the second mutate() that was being called after sign-in. So only useEffect is handling the mutate() and there's no race condition and only one notification. Using isPending didn't work since it keeps unnecessarily triggering the useEffect (and spamming mutate) and showing multiple notifications until the page finishes redirecting :)

src/server/api/routers/organizations.ts
```diff
@@ -8,10 +8,82 @@ import {
...
+    const role = user.role as Role;
+    let redirectUrl = "/";
+
+    switch (role) {
+      case Role.ADMIN:
+        redirectUrl = "/admin/home";
+        break;
+
+      case Role.AGENCY: {
+        const activeOrgId = ctx.session.session.activeOrganizationId;
```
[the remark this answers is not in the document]
tanvimahal: this also doesn't apply since ctx.session consists of a session and a user so it should be ctc.session.session :) 
image

src/app/login/page.tsx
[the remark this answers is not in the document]
promatty: > @promatty I already fixed the duplicate/multiple notifications before by removing the second mutate() that was being called after sign-in. So only useEffect is handling the mutate() and there's no race condition and only one notification. Using isPending didn't work since it keeps unnecessarily triggering the useEffect (and spamming mutate) and showing multiple notifications until the page finishes redirecting :)

ohh i see, i was still seeing the duplicate notification because i was using the back arrow. i have a feeling the fix is to change router.push to router.replace, or remove mutate being part of the dependency array. please see if there's an easy fix to this, but dont bother if you need something more complex like middleware to fix it. lmkk
https://github.com/Code-the-Change-YYC/salvationarmy/blob/5c656c17a0d0c3a13424fdcfc06652a592da663d/src/app/login/page.tsx#L22
https://github.com/Code-the-Change-YYC/salvationarmy/blob/5c656c17a0d0c3a13424fdcfc06652a592da663d/src/app/login/page.tsx#L33


https://github.com/user-attachments/assets/d482d3e7-6ab3-4dde-a9ae-15a1ca711b38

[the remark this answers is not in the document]
tanvimahal: @promatty i added useRef to prevent the useEffect from calling mutate() multiple times, and changed router.push() to router.replace() to remove the login page from browser history so that the login page gets skipped whenever you press the back button. I tested everything out again, hopefully it shld be good :))

promatty (approved):
lgtm

promatty:
@burtonjong u need to approve as well

burtonjong (approved):

burtonjong:
@tanvimahal Thanks for the hard work. Feel free to merge yourself

PR #31, opened 2025-12-01
Implement logout endpoint

JustinTan-1:
/logout
logs out of current user account

## Release Notes

* **New Features**
  * Profile menu integrated into navigation bar displaying user name and email information
  * Logout option added to profile menu with automatic session sign-out and redirect to login page
  * New logout icon added to the user interface


burtonjong (changes requested):
I guess it's fair because there isn't really a mockup for figma, but this is what victoria had:

Image

I would say that it would be best to be able to click the profile component in the navbar and have this popup as a sort of attached modal type of thing.

I would say this would either be a mantine menu: https://mantine.dev/core/menu/

Or it would be a hovercard: https://mantine.dev/core/hover-card/

I personally would like to just have the menu.

There, you can now use the authClient to signout on button press.

JustinTan-1:
Screenshot from 2025-12-01 10-49-47

src/app/logout/page.tsx
burtonjong, line 1: Do we still need this page?

src/app/_components/common/profile/profile.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <Menu shadow="md" width={250}>
+      <Menu.Target>
+        <ActionIcon variant="transparent" style={{ padding: 0, zIndex: 1 }}>
+          <User />
+        </ActionIcon>
+      </Menu.Target>
+
+      <Menu.Dropdown>
+        <Menu.Label>Name: {session?.user.name}</Menu.Label>
```
burtonjong, line 31: I'm not at home so I can't see the implementation of useSession(), but what does useSession() actually return? I thought it was whatever was in the session table, and didn't include this `user` object...? I just want to make sure.

src/app/_components/common/profile/profile.tsx
```diff
@@ -0,0 +1,34 @@
...
+export default function Profile() {
+  const { data: session } = useSession();
+  const router = useRouter();
+  function logout() {
+    router.push("/logout");
+  }
+  return (
+    <Menu shadow="md" width={250}>
+      <Menu.Target>
+        <ActionIcon variant="transparent" style={{ padding: 0, zIndex: 1 }}>
```
burtonjong: We have a shared component for this. Use that instead

burtonjong (changes requested):

src/app/_components/common/profile/profile.tsx
JustinTan-1: Note: I needed to refactor the shared component to pass ref

src/app/_components/common/profile/profile.tsx
JustinTan-1: yes it seems like betterauth has the info included in the table by default
image

burtonjong: Just surprised that useSession() returns the entire user. I thought it was just tokens/expiry dates. If it works it works lol!

burtonjong (approved):

PR #32, opened 2025-12-02
SANC-45: connect calendar view to trip data api

themaxboucher:
https://codethechangeyyc.atlassian.net/browse/SANC-45

Connected the calendar view component to the bookings from the database with proper loading and error handling. 

https://github.com/user-attachments/assets/e120ac23-dea6-4331-b5dc-36b77efc6991

## Summary by CodeRabbit

* **Improvements**
  * Shows a centered loading indicator while bookings load and a centered error alert on failure.
  * Calendar now uses live booking data (demo/sample data removed) for accurate scheduling.
  * Day headers and overall calendar layout refreshed for improved visual consistency and responsiveness.
  * Loading/error containers are vertically centered with a fixed height for predictable display.


src/app/_components/calendar-view.tsx
```diff
@@ -49,265 +49,37 @@ function getEventColor(startDate: string): string {
...
-    driverId: "driver-17",
-    purpose: "Large donation - multiple items",
-    start: "2025-11-14T13:30:00",
-    end: "2025-11-14T15:00:00",
-  },
-];
-
 interface CalendarViewProps {
-  bookings?: CalendarBooking[];
+  bookings: (typeof bookings.$inferSelect)[];
```
burtonjong: I would rather create a type for this instead of defining it here because it definitely will be used in multiple places

src/app/_components/calendar-view.tsx
```diff
@@ -49,265 +49,37 @@ function getEventColor(startDate: string): string {
...
+      purpose: booking.purpose ?? undefined,
       passengerInfo: booking.passengerInfo,
-      status: booking.status,
+      status: booking.status as BookingStatus,
       agencyId: booking.agencyId,
-      driverId: booking.driverId,
+      driverId: booking.driverId ?? undefined,
       createdAt: booking.createdAt,
-      updatedAt: booking.updatedAt,
+      updatedAt: booking.updatedAt ?? undefined,
```
burtonjong: Are we using ?? here because it's defined in CalendarEvent that any Record: <string, string | undefined>? I feel like most of thest props are required

Even if not, it might be better for some like the purpose to be an empty string instead of undefined

burtonjong: Basically, lets just confirm that some of these are required or not. Like when a booking is created, hypothetically, the createdAt will exist, but in `types.ts` we define it as an optional prop.

It might be better to define extendedProps as a Partial<Booking> if it doesn't contain all of the fields that booking the booking actually takes. Let me know if this is confusing

src/app/_components/agencycomponents/agency-interactive-area.module.scss
```diff
@@ -2,3 +2,10 @@
   display: flex;
   justify-content: center;
 }
+
+.loadingContainer, .errorContainer {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 50vh;
```
burtonjong: is it possible that the height of the loading container can just be a percentage like 100%? Is 50vh a magic number that just works on all viewport widths?

burtonjong (changes requested):

src/app/_components/calendar-view.tsx
themaxboucher: Got it

src/app/_components/calendar-view.tsx
themaxboucher: The thing is some of the fields are nullable in `booking-schema.ts` file defining the bookings table. Should we make these NOT NULL or handle null values on the frontend?

Also I'm a bit confused about the `BookingStatus` type. Should `BookingStatus` be an enum like in `types.ts` or string literal type like in `booking-schema.ts`?

src/app/_components/agencycomponents/agency-interactive-area.module.scss
themaxboucher: Good point. 50vh doesn't have any meaning to it. If we use 100% then the container only takes up the height of the loader. I changed the height to 500px so that the loader is at the centre of where the calendar will be.

**With 100%**
Screenshot 2025-12-05 at 10 43 10 AM

**With 500px**
Screenshot 2025-12-05 at 10 44 33 AM

burtonjong: I kind of like 500px better here cause its in the middle. Is there a way to not set a pixel value but still have it in the middle of the screen?

src/app/_components/calendar-view.tsx
burtonjong: You can infer the type from booking schema i believe, and then make another type that is Partial<of the type you just made> and then assign it to the extended props

themaxboucher: The `Booking` type is currently inferred from the booking schema. How should we deal with values in a booking that are null? Should we make them undefined or just make them empty strings? For example, if `booking.driverId` is `null` how should we handle that on the front-end?

src/app/_components/agencycomponents/agency-interactive-area.module.scss
themaxboucher: In that case, the only thing I can think of is going back to viewport units. We could also place the loader in the centre of the entire screen with absolute positioning, regardless of where the calendar/table view will be, but that probably wouldn't make sense.

promatty (changes requested):
in calendar-view.tsx, when we click on an entry to view details, make it a modal instead of an alert. 
also ensure the mobile view of the calendar matches the figma

themaxboucher:
I didn't add the modal since I couldn't find any designs for that modal on Figma. Should I just design it myself? 

The issue with the responsiveness on mobile is not a problem with the calendar, it's a problem with the view controller component. Should I make that component responsive in this ticket, or in a different one? 

Screenshot 2025-12-07 at 12 17 21 PM

promatty:
> I didn't add the modal since I couldn't find any designs for that modal on Figma. Should I just design it myself?
> 
> The issue with the responsiveness on mobile is not a problem with the calendar, it's a problem with the view controller component. Should I make that component responsive in this ticket, or in a different one?
> 
> Screenshot 2025-12-07 at 12 17 21 PM

oh i see, ill ask if Victoria can design us a modal for it. that will come in the future ticket then. 

dont worry about the responsiveness then, ill make a separate ticket for it as well

src/app/_components/calendar-view.tsx
burtonjong: I think there will only ever be one driver so there's no actual need to do worry about this

We can just have a conditional on the frontend instead that says if null just don't show anything at all.

burtonjong:
@themaxboucher is this good to go then?

src/app/_components/calendar-view.tsx
themaxboucher: I updated the `CalendarEvent` interface to allow for null types without converting them to be undefined. Is this what you were thinking of? If so, and you don't have any other revisions, this ticket should be good to go.

burtonjong (approved):

promatty (approved):
lgtm

PR #33, opened 2025-12-08
Implement logout page and refactor logout menu

JustinTan-1:
image
image

Layouts are fine on Mobile as well. 

TODO: pfp right now is a placeholder

## Summary by CodeRabbit

* **New Features**
  * Logout confirmation page with icons and a full-width action button that redirects.
  * Redesigned profile menu showing an integrated profile card (avatar, name, email) and a full-width logout button.
  * New header and UI icon/image components added.

* **Style**
  * Responsive styles for profile and logout UIs: card layout, typography, spacing, avatar sizing, and small-viewport adjustments.


src/app/_components/common/profile/Profile.module.scss
```diff
@@ -0,0 +1,90 @@
...
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  padding: 16px;
+  gap: 10px;
+  max-width: 90vw;
+
+  /* White */
+  background: #ffffff;
```
burtonjong: You can use what we have defined in variables.scss

src/app/_components/common/profile/Profile.module.scss
```diff
@@ -0,0 +1,90 @@
...
+
+  /* Heading 3 */
+  font-family: "Albert Sans";
+  font-style: normal;
+  font-weight: 700;
+  line-height: 24px;
+  /* identical to box height */
+
+  /* Dark Grey */
+  color: #424242;
```
burtonjong: $color-text-primary: $color-dark-grey; should be used

src/app/_components/common/profile/Profile.module.scss
```diff
@@ -0,0 +1,90 @@
...
+  box-shadow: 0px 0px 10px rgba(0, 0, 0, 0.25);
+  border-radius: 8px;
+  overflow-x: hidden;
+}
+
+.usernameText {
+  /* User name */
+
+  /* Heading 3 */
+  font-family: "Albert Sans";
```
burtonjong: There are a couple things in headings/text that we shouldn't need to define like font family or weight because these should already be defined in globals.scss

If you're overriding a certain style, that's fine, but there's no need to redefine it

src/app/_components/common/profile/Profile.module.scss
```diff
@@ -0,0 +1,90 @@
...
+}
+
+.emailText {
+  /* useremail@email.com */
+
+  width: 289px;
+  height: 15px;
+
+  /* Regular Copy */
+  font-family: "Albert Sans";
```
burtonjong: Same thing here

src/app/_components/common/profile/Profile.module.scss
burtonjong, line 1: For this file I didn't comment on everything, just clean the styles for fonts and colors up a bit to use variables and what we already have defined in variables.scss and globals.scss

src/app/_components/common/profile/profile.tsx
```diff
@@ -14,27 +16,34 @@ export default function Profile() {
     authClient
       .signOut()
       .then(() => {
-        router.push("/login");
+        router.push("/logout");
       })
       .catch((err) => {
         console.error("Logout failed:", err);
         alert("An error occurred during sign out");
```
burtonjong: Yeah you should import `notify` here and use that instead of an alert

src/app/logout/Logout.module.scss
```diff
@@ -0,0 +1,104 @@
...
+  flex-direction: column;
+  align-items: flex-start;
+  padding: 32px;
+  gap: 12px;
+
+  position: absolute;
+  width: 526px;
+  height: 297px;
+  left: calc(50% - 526px / 2);
+  top: calc(50% - 297px / 2 + 0.5px);
```
burtonjong: I really want to know what this looks like on mobile LOL

Can you explain why these styles are necessary...? I would imagine that what mantine provides is enough without having to style it like this

src/app/logout/Logout.module.scss
burtonjong, line 1: same issues here with font sizes and color vars

src/app/logout/Logout.module.scss
burtonjong: Like here: https://mantine.dev/core/menu/#dropdown-position

if you set to `bottom-end` and adjust the offset, you should be able to achieve what you do here (maybe except the width/height)

Honestly the width and height being pixel values throws me off as well. I would like to see how this looks on mobile since some screen sizes themselves are less than 500px.

burtonjong (changes requested):

src/app/logout/Logout.module.scss
JustinTan-1: oh yeah, I'm not sure if it is optimal it is copied from the Figma

JustinTan-1: Basically, I copied the Figma CSS styles for both the mobile and the desktop version then wrapped it with @media

JustinTan-1:
image
This is mobile view

src/app/logout/Logout.module.scss
JustinTan-1: I removed it now

src/app/_components/common/profile/Profile.module.scss
```diff
@@ -0,0 +1,90 @@
...
+  box-shadow: 0px 0px 10px rgba(0, 0, 0, 0.25);
+  border-radius: 8px;
+  overflow-x: hidden;
+}
+
+.usernameText {
+  /* User name */
+
+  /* Heading 3 */
+  font-style: normal;
```
burtonjong: I don't think this is needed still, it will be normal by default

src/app/_components/common/profile/Profile.module.scss
```diff
@@ -0,0 +1,90 @@
...
+  order: 0;
+  align-self: stretch;
+  flex-grow: 0;
+}
+
+.emailText {
+  /* useremail@email.com */
+
+  width: 289px;
+  height: 15px;
```
burtonjong: do we need to define an width and a height here still?

src/app/_components/common/profile/Profile.module.scss
```diff
@@ -0,0 +1,90 @@
...
+
+  /* Regular Copy */
+  font-style: normal;
```
burtonjong: Same thing as above

src/app/_components/common/profile/Profile.module.scss
```diff
@@ -0,0 +1,90 @@
...
+  order: 0;
+  align-self: stretch;
+  flex-grow: 0;
+}
+
+.profilePicture {
+  /* Ellipse 1 */
+
+  width: 48px;
+  height: 48px;
```
burtonjong: This won't grow/shrink on mobile, does it look too big on mobile though?

src/app/logout/Logout.module.scss
```diff
@@ -0,0 +1,89 @@
...
+  text-align: center;
+
+  /* Dark Grey */
+  color: $color-dark-grey;
+
+  @media (max-width: 420px) {
+    /* You’ve logged out of the Navigation Centre */
+
+    width: 289px;
+    height: 58px;
```
burtonjong: defining width and height for a text element to me seems unnecessary, it should just be based off the font size

src/app/logout/Logout.module.scss
```diff
@@ -0,0 +1,89 @@
...
+    box-shadow: 0px 0px 10px rgba(0, 0, 0, 0.25);
+    border-radius: 16px;
+  }
+}
+
+.logoutText {
+  /* You’ve logged out of the Navigation Centre */
+
+  width: 450px;
+  height: 76px;
```
burtonjong: Same here

src/app/logout/Logout.module.scss
```diff
@@ -0,0 +1,89 @@
...
+
+  /* Auto layout */
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  padding: 32px;
+  gap: 12px;
+
+  width: 526px;
+  height: 297px;
```
burtonjong: I still don't like how this is a fixed pixel width. Changing this to a percent is probably better.

burtonjong (changes requested):

JustinTan-1:
image

src/app/_components/common/profile/profile.tsx
```diff
@@ -14,27 +18,43 @@ export default function Profile() {
...
+        notifications.show({
+          id: "logout-error",
+          position: "bottom-center",
+          withCloseButton: true,
+          autoClose: 5000,
+          title: "Error",
+          message: "Logout Failed",
+          color: "white",
+          icon: <Cross />,
+        });
```
burtonjong: This should be using our shared notify component, not directly from mantine

src/app/logout/Logout.module.scss
```diff
@@ -0,0 +1,60 @@
...
+  /* Auto layout */
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  padding: 32px;
+  gap: 12px;
+
+  width: 27%;
+  min-width: 353px;
+  min-height: 279px;
```
burtonjong: I mean there must be a reason for this... ill allow it but like I still don't know why this is here

burtonjong (changes requested):

src/app/logout/Logout.module.scss
JustinTan-1: My reasoning is because if I shrink the screen really small the box will shrink to a really small size as well

JustinTan-1: lmk if there is a better solution to this

burtonjong: Ah, if that's the case I would rather have fixed sizes for this container based on media querys

So like mobile gets a set width, and tablets get a set width, smaller laptop screens get set width, etc.

We should have the queries that we use defined somewhere

burtonjong: Or it doesn't look like they're defined but I would just go with the standard:
image

JustinTan-1:
image
Ok, this is the current iteration

burtonjong (approved):

PR #34, opened 2025-12-09
SANC-43-Connect-agency-booking-form-to-trip-creation-API

wesleylui:
The agency booking form now uses the trip creation API through a TanStack Query Mutation, this also takes care of loading/error states, and success notifications. Form values are mapped to the API format with proper validation, including Google Maps address verification and datetime conversion. The form automatically resets after successful submission.
Key change:
- Connected form submission to `api.trip.create` endpoint

## Summary by CodeRabbit

* **New Features**
  * Google Places address autocomplete for bookings (CA-restricted).
  * Booking submission now uses mutation-driven loading and includes explicit phone number support.
  * New input validation helpers for string length and time-range checks.

* **Bug Fixes**
  * Expanded, clearer validation for names, phone, addresses, and times.
  * Improved date/time picker handling and stricter NaN/date validation.

* **Chores**
  * Removed an unused import; minor API/prop tightening and small cleanup.

### DIFF
--- a/src/app/_components/agencycomponents/agency-form.tsx
+++ b/src/app/_components/agencycomponents/agency-form.tsx
@@ -3,6 +3,7 @@
 import { Box, Divider, Stack, Textarea, TextInput } from "@mantine/core";
 import { DateTimePicker } from "@mantine/dates";
 import type { UseFormReturnType } from "@mantine/form";
+import dayjs from "dayjs";
 import classes from "./agency-form.module.scss";
 
 interface AgencyBookingForm {
@@ -24,30 +25,6 @@ interface AgencyFormProps {
 
 export const AgencyForm = ({ form, destinationAddressRef }: AgencyFormProps) => {
   const now = new Date();
-
-  // Helper function to convert Date or string to local ISO string (preserves local time, no UTC conversion)
-  const toLocalISOString = (value: Date | string | null): string => {
-    if (!value) return "";
-
-    // Normalize input to Date object
-    let date: Date;
-    if (typeof value === "string") {
-      date = new Date(value);
-      // Check if the parsed date is invalid
-      if (Number.isNaN(date.getTime())) return "";
-    } else {
-      date = value;
-    }
-
-    const year = date.getFullYear();
-    const month = String(date.getMonth() + 1).padStart(2, "0");
-    const day = String(date.getDate()).padStart(2, "0");
-    const hours = String(date.getHours()).padStart(2, "0");
-    const minutes = String(date.getMinutes()).padStart(2, "0");
-    const seconds = String(date.getSeconds()).padStart(2, "0");
-
-    return `${year}-${month}-${day}T${hours}:${minutes}:${seconds}`;
-  };
 
   return (
     <Stack gap="lg">
@@ -109,14 +86,24 @@ export const AgencyForm = ({ form, destinationAddressRef }: AgencyFormProps) => 
             label="Start time"
             placeholder="Select start date and time"
             minDate={now}
-            valueFormat="YYYY-MM-DDTHH:mm:ss"
+            valueFormat="MMM DD, YYYY hh:mm A"
             value={form.values.startTime ? new Date(form.values.startTime) : null}
             onChange={(value) => {
-              const localISOString = toLocalISOString(value);
-              form.setFieldValue("startTime", localISOString);
+              if (!value) {
+                form.setFieldValue("startTime", "");
+                return;
+              }
+              const isoString = dayjs(value).toISOString();
+              console.log("Start Time - parsed:", isoString);
+              console.log("Start Time - after parsing:", new Date(isoString));
+              console.log(
+                "Start Time - show local time:",
+                dayjs(isoString).format("YYYY-MM-DD hh:mm A"),
+              );
+              form.setFieldValue("startTime", isoString);
 
-              // to reset the end time if it's invalid
-              if (localISOString && form.values.endTime && form.values.endTime <= localISOString) {
+              if (isoString && form.values.endTime && form.values.endTime <= isoString) {
+                // to reset the end time if it's invalid
                 form.setFieldValue("endTime", "");
               }
             }}
@@ -135,11 +122,21 @@ export const AgencyForm = ({ form, destinationAddressRef }: AgencyFormProps) => 
             label="End time"
             placeholder="Select end date and time"
             minDate={form.values.startTime ? new Date(form.values.startTime) : now}
-            valueFormat="YYYY-MM-DDTHH:mm:ss"
+            valueFormat="MMM DD, YYYY hh:mm A"
             value={form.values.endTime ? new Date(form.values.endTime) : null}
             onChange={(value) => {
-              const localISOString = toLocalISOString(value);
-              form.setFieldValue("endTime", localISOString);
+              if (!value) {
+                form.setFieldValue("endTime", "");
+                return;
+              }
+              const isoString = dayjs(value).toISOString();
+              console.log("End Time - parsed:", isoString);
+              console.log("End Time - after parsing:", new Date(isoString));
+              console.log(
+                "End Time - show local time:",
+                dayjs(isoString).format("YYYY-MM-DD hh:mm A"),
+              );
+              form.setFieldValue("endTime", isoString);
             }}
             disabled={!form.values.startTime}
             timePickerProps={{
--- a/src/app/_components/agencycomponents/agency-form.tsx
+++ b/src/app/_components/agencycomponents/agency-form.tsx
@@ -25,9 +25,19 @@ interface AgencyFormProps {
 export const AgencyForm = ({ form, destinationAddressRef }: AgencyFormProps) => {
   const now = new Date();
 
-  // Helper function to convert Date to local ISO string (preserves local time, no UTC conversion)
-  const toLocalISOString = (date: Date | null): string => {
-    if (!date) return "";
+  // Helper function to convert Date or string to local ISO string (preserves local time, no UTC conversion)
+  const toLocalISOString = (value: Date | string | null): string => {
+    if (!value) return "";
+
+    // Normalize input to Date object
+    let date: Date;
+    if (typeof value === "string") {
+      date = new Date(value);
+      // Check if the parsed date is invalid
+      if (Number.isNaN(date.getTime())) return "";
+    } else {
+      date = value;
+    }
 
     const year = date.getFullYear();
     const month = String(date.getMonth() + 1).padStart(2, "0");
--- a/src/app/_components/agencycomponents/agency-form.tsx
+++ b/src/app/_components/agencycomponents/agency-form.tsx
@@ -25,6 +25,20 @@ interface AgencyFormProps {
 export const AgencyForm = ({ form, destinationAddressRef }: AgencyFormProps) => {
   const now = new Date();
 
+  // Helper function to convert Date to local ISO string (preserves local time, no UTC conversion)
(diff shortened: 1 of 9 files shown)


burtonjong (commented):
@wesleylui I think as part of this ticket I wouldn't mind cleaning up some of the validation on the frontend for this.

Things like:
- residentName shouldn't be crazy long (or any of the fields for that matter, maybe except the purpose of the trip)
-  endTime shouldn't be less that startTime

If you think there's some validation function that is reusable, you can also put that in validation.ts or whatever the shared validation file is called (has the email validation there as well)

wesleylui:
Added reusable validation utilities to `validation.ts`:

- **`validateStringLength()`**: General-purpose string validation with min/max length constraints and custom field names for error messages
- **`validateTimeRange()`**: Validates that end time is after start time for datetime fields

These validators are used in the agency booking form but can be reused in other forms for validating general strings and dates

src/app/_components/agencycomponents/agency-interactive-area.tsx
```diff
@@ -55,15 +67,24 @@ export const BookingInteractiveArea = ({ initialViewMode = ViewMode.CALENDAR }:
...
-      purpose: (value) => (value.trim().length > 0 ? null : "Purpose is required"),
-      pickupAddress: (value) => (value.trim().length > 0 ? null : "Pickup address is required"),
-      destinationAddress: (value) =>
-        value.trim().length > 0 ? null : "Destination address is required",
+      endTime: (value, values) => {
+        // First check if required
+        if (value.trim().length === 0) return "Date and time is required";
+        // Then validate time range
+        return validateTimeRange(values.startTime, value);
+      },
```
promatty, line 98: the agency-form only displays a single datepicker for date and time of transport. have it show a datepicker for both the start and end times instead of a transportDateTime

src/app/_components/agencycomponents/agency-interactive-area.tsx
```diff
@@ -114,54 +135,31 @@ export const BookingInteractiveArea = ({ initialViewMode = ViewMode.CALENDAR }:
...
-      notify.success("Booking successfully created");
-      form.reset();
-    }, 2000);
+    createBookingMutation.mutate({
+      title: form.values.title,
+      residentName: form.values.residentName,
+      contactInfo: form.values.contactInfo,
+      additionalInfo: form.values.additionalInfo,
+      pickupAddress: form.values.pickupAddress,
+      destinationAddress: inputElement.current?.value || "",
```
promatty, line 185: why do we use the ref value instead of form value?

src/app/_components/agencycomponents/agency-interactive-area.tsx
```diff
@@ -23,12 +25,22 @@ const GOOGLE_MAPS_LIBRARIES_ARRAY: Libraries = ["places"]; //Add more to this ar
...
   const [currentDate, setCurrentDate] = useState<Date>(new Date());
   const [isDayView, setIsDayView] = useState<boolean>(false);
   const [validationAddressGood, setValidationAddressGood] = useState<boolean>(false);
 
+  const createBookingMutation = api.trip.create.useMutation({
+    onSuccess: () => {
+      notify.success("Booking successfully created");
+      form.reset();
+      setShowBookingModal(false);
+      setValidationAddressGood(false);
```
promatty, line 44: after creating a new booking, the table and calendar wont refresh. perhaps we should invalidate the bookings query to cause a refresh
```suggestion
      setValidationAddressGood(false);
      void utils.bookings.invalidate();
```

src/types/validation.ts
```diff
@@ -8,3 +8,52 @@ export const passwordSchema = z
...
+
+  const start = new Date(startTime);
+  const end = new Date(endTime);
+
+  if (end <= start) {
+    return "End time must be after start time";
+  }
+
+  return null;
+};
```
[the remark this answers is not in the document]
promatty: yeah lets get some validation on whether the date is valid

promatty (changes requested):
also creation doesnt work rn since the datepicker doesnt work. this is likely an issue with iso strings not converting to a Date properly, but may also be how we no longer use this transportDateTime field 

https://github.com/user-attachments/assets/54a8e977-1a52-431e-bd76-87add82f787c

src/app/_components/agencycomponents/agency-interactive-area.tsx
wesleylui: I believe we have to use the ref value for destinationAddress because of the Google autocomplete. I believe it'll autocomplete the value directly in the HTML element so we will get the full address from there and not the original input on the form.

burtonjong (commented):
@wesleylui 

You'll have to take another look at https://github.com/Code-the-Change-YYC/salvationarmy/pull/40/changes#diff-09231f59b104026943ace40134be0e25e8f2821eb96983bf8b5e02721e2834dd

once it's merged.

You might have to merge that into your branch and fix some things. 

You have to be careful with the start/end times, as there are many things that can go wrong with timestamps

Just for example:
- Drizzle declares them in booking schema as below, but what does that really mean?
```
  startTime: timestamp("start_time", {
    mode: "string",
    withTimezone: true,
  }).notNull(),
```

- Supabase stores timestamps as `timestamptz`
- `Date.prototype.toISOString()` always converts the date to UTC in the standard ISO 8601 format, so that could cause problems 
    -  basically when the agency books the booking, we have to make sure when it's reflected on the schedule that it is in LOCAL TIME, not UTC

burtonjong:
Here's what I mean: {
    "title": "Test",
    "residentName": "Test",
    "phoneNumber": "Test",
    "additionalInfo": "Test",
    "startTime": "2026-01-08 12:43:00",
    "endTime": "2026-01-08 20:42:00",
    "purpose": "Hi!",
    "pickupAddress": "asdf",
    "destinationAddress": "asdf"
}

it is sent like this. I don't even think this can be parsed into a valid format. You might have to change how this date is stored!

burtonjong:
You should be able to format it the way you want to change it to an isostring with the valueFormat prop:

https://mantine.dev/dates/date-time-picker/?t=props
image

wesleylui:
Changes:
- Removed an extra .toISOString() conversion in the handleconfirm() of the agency-interactive-area extra that was causing the local times to convert to UTC.
- Dates are now passed directly from form values (already in local ISO format)
- Also added a value format of "YYYY-MM-DDTHH:mm:ss" to the DateTimePicker start and end times

Lmk if this makes sense, I am a bit confused on the Date stuff

burtonjong:
@wesleylui Not sure when I can take another look at this, but I have seen this. I need to test it myself

src/app/_components/agencycomponents/agency-form.tsx
```diff
@@ -85,12 +109,14 @@ export const AgencyForm = ({ form, destinationAddressRef }: AgencyFormProps) =>
             label="Start time"
             placeholder="Select start date and time"
             minDate={now}
-            value={form.values.startTime}
+            valueFormat="YYYY-MM-DDTHH:mm:ss"
```
burtonjong:
```
 valueFormat="MMM DD, YYYY hh:mm A"
```

For readability this should be like this

burtonjong:
@wesleylui Maybe what I said was confusing lol

You can do something like this with dayjs (which is what mantine uses): 

```
const parsed = dayjs(value).toISOString();

 console.log("parsed:", parsed);

console.log("after parsing:", new Date(parsed));

console.log("show local time:", dayjs(parsed).format("YYYY-MM-DD hh:mm A"));
```

This will output this if this:
```
parsed: 2026-01-11T22:05:00.000Z

after parsing: Sun Jan 11 2026 15:05:00 GMT-0700 (Mountain Standard Time)

show local time: 2026-01-11 03:05 PM
```

We should be storing it as an ISO string.

burtonjong (changes requested):

burtonjong:
> @wesleylui Maybe what I said was confusing lol
> 
> You can do something like this with dayjs (which is what mantine uses):
> 
> ```
> const parsed = dayjs(value).toISOString();
> 
>  console.log("parsed:", parsed);
> 
> console.log("after parsing:", new Date(parsed));
> 
> console.log("show local time:", dayjs(parsed).format("YYYY-MM-DD hh:mm A"));
> ```
> 
> This will output this if this:
> 
> ```
> parsed: 2026-01-11T22:05:00.000Z
> 
> after parsing: Sun Jan 11 2026 15:05:00 GMT-0700 (Mountain Standard Time)
> 
> show local time: 2026-01-11 03:05 PM
> ```
> 
> We should be storing it as an ISO string.

What I am saying here is that you don't need to create your own helper to change to isostring, you can just use dayjs to do the formatting and stuff.

I guess chatGPt explains the best:

```
what is the flow for how dates should be handled and then stored from frontend to backend, back to fetching from backend to frontend?
```

```
Pick time

User selects a date and time in the UI (ex: Jan 11, 2026 at 2:44 PM).
This is a human / wall-clock time and does not yet represent a real moment in time.

At this point, it’s just “what the user picked”, not something safe to store.

Normalize

Frontend converts that picked time into one agreed-upon timezone (UTC or a fixed business timezone like MST).

Example (fixed MST):

2026-01-11T14:44:00-07:00


This step removes ambiguity and prevents timezone & daylight-saving bugs.

Send ISO

Frontend sends only an ISO 8601 datetime string to the backend.

{
  "startsAt": "2026-01-11T14:44:00-07:00"
}


ISO strings are machine-safe, language-agnostic, and sortable.

Store

Backend validates the datetime and stores it as a real instant in time.

With timezone info

Not as a formatted string

Not as local time

Storage should always be unambiguous and consistent.

Return ISO

Backend returns the same ISO datetime format when data is fetched.

2026-01-11T14:44:00-07:00


APIs should never return human-formatted dates.

Display nicely

Frontend converts the ISO string back into a Date and formats it for humans.

Example:

Jan 11, 2026 — 2:44 PM


Formatting is only for display — never for storage or APIs.
```

src/app/_components/agencycomponents/agency-form.tsx
```diff
@@ -109,14 +86,24 @@ export const AgencyForm = ({ form, destinationAddressRef }: AgencyFormProps) =>
...
+                form.setFieldValue("startTime", "");
+                return;
+              }
+              const isoString = dayjs(value).toISOString();
+              console.log("Start Time - parsed:", isoString);
+              console.log("Start Time - after parsing:", new Date(isoString));
+              console.log(
+                "Start Time - show local time:",
+                dayjs(isoString).format("YYYY-MM-DD hh:mm A"),
+              );
```
burtonjong: Delete the console logs

src/app/_components/agencycomponents/agency-form.tsx
```diff
@@ -135,11 +122,21 @@ export const AgencyForm = ({ form, destinationAddressRef }: AgencyFormProps) =>
...
+                form.setFieldValue("endTime", "");
+                return;
+              }
+              const isoString = dayjs(value).toISOString();
+              console.log("End Time - parsed:", isoString);
+              console.log("End Time - after parsing:", new Date(isoString));
+              console.log(
+                "End Time - show local time:",
+                dayjs(isoString).format("YYYY-MM-DD hh:mm A"),
+              );
```
burtonjong: Same as above

burtonjong (changes requested):

burtonjong:
@wesleylui address comments and fix merge conflicts then this look sgood

wesleylui:
should be ready to merge :)

burtonjong (approved):

promatty (approved):
ts tuff 🔥

PR #36, opened 2025-12-24
SANC-48-When-the-user-completes-registration-they-should-be-able-to-input-their-name

jason-duong4509:
- Completed SANC-48
- Added a loading screen while the Google Maps API loads (feedback from SANC-39). Found in \src\app\_components\agencycomponents\agency-interactive-area.tsx

Attached is a video demonstrating the login process, starting when the user has a name (directs them to /dashboard) and then when the user has no name (directs them to /fill-out-name).

In the video, /dashboard was a copy-paste of the agency-interative-area.tsx file. It was only added since the endpoint does not yet exist. It is not included in the final branch (/dashboard remains non-existent)

https://github.com/user-attachments/assets/3abf4830-181c-4425-a4d6-155bf999d5c8

## Summary by CodeRabbit

* **New Features**
  * Added a "Fill Out Name" page for users to enter and save their profile name and be redirected after saving.
  * New server endpoint to change a user's name and support the post-save redirect flow.

* **Improvements**
  * Users without a name are routed to the name entry page before dashboard access.
  * Loading now shows a styled, centered loading card instead of an empty placeholder.

* **Validation**
  * Name input enforces format rules to prevent invalid entries.


src/app/fill-out-name/page.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  const handleSubmit = async (values: typeof form.values) => {
+    setLoading(true);
+    //Alphanumeric regex of length >=3. Allows for exactly 1 space between characters
+    const regex = /^(?!.*\s{2})[A-Za-z0-9][A-Za-z0-9\s]*[A-Za-z0-9][A-Za-z0-9\s]*[A-Za-z0-9]$/;
+    if (!regex.test(values.name)) {
+      //User inputted name is not proper
+      notify.error("Invalid Name");
+      setLoading(false);
+    } else {
```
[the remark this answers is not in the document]
jason-duong4509: 3 character names do work. Resolving conversation as a result

src/server/api/routers/organizations.ts
```diff
@@ -224,4 +226,34 @@ export const organizationRouter = createTRPCRouter({
...
+        .returning(); //Make DB update command to change the user's name
+
+      if (listOfUpdatedUsers.length === 0) {
+        //Drizzle updated no users (something went wrong)
+        throw new TRPCError({
+          code: "INTERNAL_SERVER_ERROR",
+          message: "Failed to update user's name",
+        });
+      }
+    }),
```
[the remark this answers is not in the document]
jason-duong4509: 1. done
2. likely not needed 
3. does work for 3 characters

src/app/fill-out-name/page.tsx
```diff
@@ -0,0 +1,75 @@
...
+    onError: (error) => {
+      notify.error(error.message);
+      setLoading(false);
+    },
+  });
+
+  const handleSubmit = async (values: typeof form.values) => {
+    setLoading(true);
+    //Alphanumeric regex of length >=3. Allows for exactly 1 space between characters
+    const regex = /^(?!.*\s{2})[A-Za-z0-9][A-Za-z0-9\s]*[A-Za-z0-9][A-Za-z0-9\s]*[A-Za-z0-9]$/;
```
burtonjong: You should be able to add this validation to `validation.ts` and then move the actual validation checking to the useForm() call (because zod will handle it).

src/app/fill-out-name/page.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const [loading, setLoading] = useState(false);
+  const router = useRouter();
+
+  const form = useForm({
+    initialValues: {
+      name: "",
+    },
+  });
+
+  const changeNameMutation = api.organization.changeName.useMutation({
```
burtonjong, line 32: If I'm not mistaken, you should be able to extract a loading state from this mutation instead of making your own with useState().

src/server/api/routers/organizations.ts
```diff
@@ -224,4 +226,34 @@ export const organizationRouter = createTRPCRouter({
...
+      const regex = /^(?!.*\s{2})[A-Za-z0-9][A-Za-z0-9\s]*[A-Za-z0-9][A-Za-z0-9\s]*[A-Za-z0-9]$/;
+      if (!regex.test(newName)) {
+        //User inputted name is not proper
+        throw new TRPCError({
+          code: "BAD_REQUEST",
+          message: "Invalid Name",
+        });
+      }
+
+      const listOfUpdatedUsers = await ctx.db
```
burtonjong, line 321: list of updated users? If you only update the user where `user.id` === `ctx.session.user.id` shouldn't there only be one anyway?

src/lib/auth.ts
```diff
@@ -11,6 +12,19 @@ export const auth = betterAuth({
...
   }),
+  hooks: {
+    after: createAuthMiddleware(async (ctx) => {
+      if (ctx.path.startsWith("/sign-in/email")) {
+        const accountName = ctx.context.newSession?.user.name; //Get the account name
+        if (accountName === "") {
+          //Account name not set
+          return { newLink: "/fill-out-name" };
+        }
+        //TODO V: change later when we know where to send the user
```
burtonjong: is this `return { newLink : "/dashboard" };` just how nextauth suggests that you redirect the user on a serverside call like this?

burtonjong (changes requested):

src/lib/auth.ts
jason-duong4509: No. Assuming you mean betterauth (unless they're the same thing), betterauth suggests using a throw to redirect, `throw ctx.redirect("/sign-up/name");`. However, since the login uses a fetch call, that throw will cause the server to redirect, not the browser (`result.data` ends up containing the code of `/fill-out-name` or `/dashboard` I believe). The browser does not redirect when the backend uses throw and fetch is used

From what I found looking into it, the typical way of doing it is to have the after hook return success on a good login and have the frontend redirect itself (like what was previously in the code). In this case, though, the frontend won't know whether to navigate to `/dashboard` or `/fill-out-name` since the backend determines that. I chose to have the backend return a JSON containing the endpoint to redirect to

Upon looking at it some more, I think the after hook can be eliminated entirely and the frontend can take care of where to redirect the user upon successful login. Both ways achieve the same result; I previously wasn't able to figure out a way to get the frontend to know where to redirect. I can change the code and make the frontend redirect properly if desired

src/server/api/routers/organizations.ts
jason-duong4509: The naming was just for me to remember that an `UPDATE` with `RETURNING` gives a list of all affected users. In this case, it's always a list of size 1 or 0 (if the DB could not find a user to update), like you mentioned

Later in that code snippet, `listOfUpdatedUsers` is checked to see whether the DB was able to update a user or not. That is the only purpose of that variable

src/app/fill-out-name/page.tsx
jason-duong4509: Can you tell me more about this? I am not following

promatty:
@burtonjong

src/lib/auth.ts
burtonjong: @jason-duong4509 I think I would prefer the frontend redirects, not the server. Move that functiionality out and then this looks good

src/app/fill-out-name/page.tsx
burtonjong: I believe you should be able to do `changeNameMutation.isPending` instead of having a loading useState.

src/lib/auth.ts
jason-duong4509: Done!

src/app/fill-out-name/page.tsx
jason-duong4509: Sounds good, made that change

src/app/fill-out-name/page.tsx
jason-duong4509: Done!

jason-duong4509:
@burtonjong

src/app/fill-out-name/page.tsx
```diff
@@ -20,11 +20,19 @@ export default function FillOutNamePage() {
     },
   });
 
+  const { mutate } = api.organization.redirectToDashboard.useMutation({
```
burtonjong: calling this function `mutate()` is misleading. A more descriptive name is better

src/server/api/routers/organizations.ts
```diff
@@ -31,50 +31,54 @@ export const organizationRouter = createTRPCRouter({
...
+          redirectUrl = "/admin/home";
+          break;
+
+        case Role.AGENCY: {
+          const activeOrgId = ctx.session.session.activeOrganizationId;
+
+          if (!activeOrgId) {
+            throw new TRPCError({
+              code: "NOT_FOUND",
+              message: "No active organization found for this agency user",
```
burtonjong: Maybe add to this at the end: please contact an administator for help... something like that

src/server/api/routers/organizations.ts
```diff
@@ -31,50 +31,54 @@ export const organizationRouter = createTRPCRouter({
...
+            throw new TRPCError({
+              code: "NOT_FOUND",
+              message: "Organization not found",
+            });
+          }
+
+          if (!organization.slug) {
+            throw new TRPCError({
+              code: "INTERNAL_SERVER_ERROR",
+              message: "Organization is missing required slug configuration",
```
burtonjong: Same thing here as above

burtonjong (changes requested):

src/server/api/routers/organizations.ts
jason-duong4509: Done!

src/server/api/routers/organizations.ts
jason-duong4509: Done!

src/app/fill-out-name/page.tsx
jason-duong4509: Done!

src/server/api/routers/organizations.ts
```diff
@@ -29,50 +31,56 @@ export const organizationRouter = createTRPCRouter({
...
-      case Role.AGENCY: {
-        const activeOrgId = ctx.session.session.activeOrganizationId;
-
-        if (!activeOrgId) {
-          throw new TRPCError({
-            code: "NOT_FOUND",
-            message: "No active organization found for this agency user",
+    if (ctx.session.user.name === "") {
+      redirectUrl = "/fill-out-name";
+    } else {
```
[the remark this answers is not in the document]
jason-duong4509: This does not actually happen. Resolving comment as a result

src/server/api/routers/organizations.ts
```diff
@@ -296,4 +304,32 @@ export const organizationRouter = createTRPCRouter({
...
+        .returning(); //Make DB update command to change the user's name
+
+      if (listOfUpdatedUsers.length === 0) {
+        //Drizzle updated no users (something went wrong)
+        throw new TRPCError({
+          code: "INTERNAL_SERVER_ERROR",
+          message: "Failed to update user's name",
+        });
+      }
+    }),
```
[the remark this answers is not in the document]
burtonjong: @jason-duong4509 This is probably going to end up being a problem. Updating the db user name and not updating the session user name will cause an inconsistency.

On line 20 we already get the user. We can just use that user instead of ctx.session.user.name, unless you thinbk otherwise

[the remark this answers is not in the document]
jason-duong4509: Done!

burtonjong (approved):

PR #39, opened 2026-01-01
Sanc 59 vehicle logs schema

JustinTan-1:
## Summary by CodeRabbit

* **New Features**
  * Vehicle log tracking for recording driver trips and vehicle usage.
  * Automatic distance calculation from odometer readings per entry.
  * Driver logs linked to user accounts with timestamps, driver/vehicle info, and audit fields.
  * Validation enforcing odometer and time consistency.

* **Chores**
  * Vehicle-log schema integrated into the database schema and configuration.

### DIFF
--- a/src/server/db/vehicle-log.ts
+++ b/src/server/db/vehicle-log.ts
@@ -16,9 +16,11 @@ export const logs = pgTable(
     kilometersDriven: integer("kilometers_driven")
       .generatedAlwaysAs((): SQL => sql`${logs.odometerEnd} - ${logs.odometerStart}`)
       .notNull(),
-    driverId: text("driver_id").references(() => user.id, {
-      onDelete: "set null",
-    }),
+    driverId: text("driver_id")
+      .references(() => user.id, {
+        onDelete: "set null",
+      })
+      .notNull(),
     driverName: text("driver_name").notNull(),
     // Name of the vehicle
     vehicle: text("vehicle").notNull(),
--- a/src/server/db/vehicle-log.ts
+++ b/src/server/db/vehicle-log.ts
@@ -17,7 +17,7 @@ export const logs = pgTable(
       .generatedAlwaysAs((): SQL => sql`${logs.odometerEnd} - ${logs.odometerStart}`)
       .notNull(),
     driverId: text("driver_id").references(() => user.id, {
-      onDelete: "set null",
+      onDelete: "restrict",
     }),
     driverName: text("driver_name").notNull(),
     // Name of the vehicle
--- a/src/server/db/vehicle-log.ts
+++ b/src/server/db/vehicle-log.ts
@@ -31,7 +31,7 @@ export const logs = pgTable(
   },
   (table) => [
     check("odometer_check", sql`${table.odometerEnd} > ${table.odometerStart}`),
-    check("date_check", sql`${table.arrivalTime} > ${table.departureTime}`),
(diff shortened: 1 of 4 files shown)


JustinTan-1:
@coderabbitai review

src/server/db/vehicle-log.ts
```diff
@@ -0,0 +1,44 @@
...
+import { user } from "./auth-schema";
+
+export const logs = pgTable(
+  "logs",
+  {
+    id: serial("id").primaryKey(),
+    date: date("date").notNull(),
+    travelLocation: text("travel_location").notNull(),
+    departureTime: time("departure_time", { precision: 0 }).notNull(),
+    arrivalTime: time("arrival_time", { precision: 0 }).notNull(),
```
promatty: your departureTime and arrivalTime use a time type without a timezone, but bookings use a timestamp type with timezone. this may cause inconsistencies

src/server/db/vehicle-log.ts
```diff
@@ -0,0 +1,44 @@
...
+    vehicle: text("vehicle").notNull(),
+    createdAt: timestamp("created_at").defaultNow().notNull(),
+    updatedAt: timestamp("updated_at")
+      .defaultNow()
+      .$onUpdate(() => /* @__PURE__ */ new Date()),
+    createdBy: text("created_by").references(() => user.id),
+    updatedBy: text("updated_by").references(() => user.id),
+  },
+  (table) => [
+    check("odometer_check", sql`${table.odometerEnd} > ${table.odometerStart}`),
```
promatty, line 42: might be an edge case where a trip starts before midnight, but ends after midnight

promatty (commented):
also export this schema in salvationarmy/src/server/db/index.ts

src/server/db/vehicle-log.ts
JustinTan-1: Changed to timezone string

src/server/db/vehicle-log.ts
JustinTan-1: You're right, I removed this check for now

promatty: i think we should still have the check, but we just need to check the date as well. since these fields are timezone types now, do they provide a date? if not, what other ways can we include date in the check?

JustinTan-1: So actually the date comes with the timestamp. When you run a check, it will internally compare both the time and the date. I tested it with some queries and I have attached below a screenshot. I believe my previous check will work unless I am missing something.
image

src/server/db/vehicle-log.ts
```diff
@@ -0,0 +1,52 @@
...
+      .notNull(),
+    driverName: text("driver_name").notNull(),
+    // Name of the vehicle
+    vehicle: text("vehicle").notNull(),
+    createdAt: timestamp("created_at").defaultNow().notNull(),
+    updatedAt: timestamp("updated_at")
+      .defaultNow()
+      .$onUpdate(() => /* @__PURE__ */ new Date()),
+    createdBy: text("created_by").references(() => user.id),
+    updatedBy: text("updated_by").references(() => user.id),
```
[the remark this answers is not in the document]
promatty: lets add the .notnull @JustinTan-1. should be good after

promatty (approved):
lgtm! good stuff

PR #46, opened 2026-01-14
Add day calendar view for Driver

JustinTan-1:
TODO: Test on an actual mobile device

Add calendar view for Driver homepage
image

## Summary by CodeRabbit

* **New Features**
  * Calendar scheduling added to Driver Home with optional toolbar buttons for navigation.

* **Behavior**
  * Appointment slots configured from 08:00 to 19:00 with 1-hour durations.
  * Sample booking data shown in the driver calendar view.

* **Style**
  * Updated calendar header and button styling for improved spacing and clarity.


src/app/driver/home/page.tsx
```diff
@@ -21,6 +22,29 @@ export default async function DriverHome() {
...
+              //Test data seeded on January 13
+              startTime: "2026-01-13T11:00:00-05:00",
+              endTime: "2026-01-13T13:30:00-05:00",
+              driverId: "423",
+              createdAt: new Date(),
+              updatedAt: new Date(),
+              createdBy: "testUser",
+              updatedBy: "testUser",
+            },
+          ]}
```
burtonjong: just for now we can extract this to a variable instead

src/app/_components/drivercomponents/calendar-view.tsx
```diff
@@ -0,0 +1,179 @@
...
+          left: "",
+          center: "title",
+          right: "prev,next",
+        }}
+        events={events}
+        eventClick={handleEventClick}
+        eventContent={renderEventContent}
+        slotMinTime="08:00:00"
+        slotMaxTime="19:00:00"
+        slotDuration="01:00:00"
```
burtonjong: I feel like we should make strings like this constant variables somewhere in a file that are exported acrsos the repo so ALL calendar components (driver, agency and admin) can use these

burtonjong: Not only that, I believe there's even more logic we can extract here, if not reuisng the calendar components that are being used in the agency/admin pages ALREADY

burtonjong (commented):
I feel like there is a lot here that is duplicated from the existing agency calendar-view.tsx component

src/app/_components/drivercomponents/calendar-view.tsx
JustinTan-1: good point, I changed it to use the one already created by agency and passed props instead

burtonjong (changes requested):
@JustinTan-1 Think it looks good, but I would move the calendar-view component to the shared/common folder now instead because it's used by multiple pages. Thanks!

burtonjong (approved):
lgtm

promatty (changes requested):
Image

can we make the date text smaller?

Image

also padding is a bit weird on bigger screens. although we want to optimize the most for mobile, computers should look decent too

JustinTan-1:
image
image

new views

promatty (approved):
lgtm

PR #47, opened 2026-01-15
Sanc 64 update the invitations frontend

Yemyam:
Update invite user and agency modals to match Figma designs

  Frontend Changes:
  - Implemented invite modal design from figma
  - Affiliated Agency dropdown now only shows for Agency member role
  - Updated User and Home icons to accept custom stroke colors
  - Added an org role select in the user invite form

NOTE:
  - The agency invite form is currently just hooked up with ```createOrganization```
  - App role (not org role) is not being handled by ```inviteUser```, every user is being added to the db as a driver.
  

## Summary by CodeRabbit

* **New Features**
  * Three-part invite flow: select invite type, invite a user, or create an agency.
  * New Agency invite form with name and slug validation.
  * Enhanced User invite form with role selection, conditional organization assignment, and improved validations.

* **UI Improvements**
  * Increased modal footer spacing for cleaner layout.
  * Icon components updated to use a consistent, customizable stroke for visual consistency.


src/app/_components/admincomponents/admin-dashboard.tsx
```diff
@@ -3,81 +3,242 @@
...
+        notify.error("Admins organization not found");
+        return;
+      }
+      organizationId = adminsOrg.id;
+    } else if (role === OrganizationRole.OWNER) {
+      const driversOrg = organizations.data?.find((org) => org.slug === "drivers");
+      if (!driversOrg) {
+        notify.error("Drivers organization not found");
+        return;
+      }
```
burtonjong: How does this make sense...? If the organization role of the user chosen is owner we want to make the organizationid the drivers org...?

src/app/_components/admincomponents/admin-dashboard.tsx
```diff
@@ -3,81 +3,242 @@
...
+      return;
+    }
+
+    const formValues = userForm.getValues();
+
+    // Determine the organization ID based on role
+    let organizationId: string;
+    const role = formValues.organizationRole;
+
+    if (role === OrganizationRole.ADMIN) {
```
burtonjong: Maybe there's some confusion here about organization roles and general roles

Better auth has org roles that can dictate what MEMBERS of the ORGANIZATION can access, and there's the defined roles that we made (driver, admin, agency)

These two concepts are different

Ex: The agency can have members, but can also have admins that can invite users to become members

src/app/_components/admincomponents/admin-dashboard.tsx
burtonjong, line 1: In this file, what is the distinction between the invite types of "user", "agency" and null?

src/app/_components/admincomponents/invite-user-form.tsx
```diff
@@ -18,44 +18,86 @@ interface InviteUserFormProps {
 
 const NO_ORGS_DATA = [{ value: "", label: "no organizations available", disabled: true }];
 
+const ROLE_LABELS: Record<OrganizationRole, string> = {
+  [OrganizationRole.ADMIN]: "Administrator",
+  [OrganizationRole.OWNER]: "Driver",
+  [OrganizationRole.MEMBER]: "Agency member",
+};
+
```
burtonjong: I commented above about this misunderstanding. Take a look

burtonjong (commented):
I think there's some confusion about these organization roles

Maybe we can hop on a call and i can explain it better if you still don't understand from my comments.

We would use these organizations in the future for these example cases:

not everyone in the salvation army navigation center should be able to invite new agencies. This could be done by appointing ADMINs (org role) and MEMBERs (org role) that have different permissions. Admins will be able to send invitations, and members wont be able to

Or even for drivers, ADMIN drivers will be able to see ALL bookings whereas MEMBER drivers won't, and only bookings assigned to them, etc.

src/app/_components/admincomponents/admin-dashboard.tsx
Yemyam: I'm using invite type to determine which modal to display. "user" for the user modal, "agency" for the agency modal, and null for the selection modal to select between inviting a user and agency.

src/app/_components/admincomponents/admin-dashboard.tsx
```diff
@@ -3,81 +3,244 @@
...
 
     if (validation.hasErrors) {
       notify.error("Please fix the errors in the form before submitting");
       return;
     }
 
-    // todo: handle case where admin permissions are given within the organization
-    console.log("submit", form.values);
+    const formValues = agencyForm.getValues();
+    console.log("submit agency", formValues);
```
promatty: remove console log

src/app/_components/admincomponents/admin-dashboard.tsx
```diff
@@ -3,81 +3,229 @@
...
+      email: userForm.values.email,
+      organizationRole: userForm.values.organizationRole,
+    };
+
+    // Only add organizationId if the role is MEMBER
+    if (userForm.values.organizationRole === OrganizationRole.MEMBER) {
+      submitData.organizationId = userForm.values.organizationId;
+    }
+
+    console.log("submit user", submitData);
```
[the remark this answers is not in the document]
promatty: remove console log

src/app/_components/admincomponents/invite-user-form.tsx
```diff
@@ -18,44 +21,100 @@ interface InviteUserFormProps {
...
+    }
+  };
+
   return (
-    <Stack gap="lg">
+    <Stack gap="xl">
       <Stack gap="md">
-        <Box fw={500} fz="lg">
-          User Information
+        <Box fw={600} fz="lg" c="#8B2635">
```
promatty: this color is hardcoded here a few times. can we import from the global scss file instead

src/app/_components/admincomponents/invite-user-form.tsx
```diff
@@ -18,44 +21,100 @@ interface InviteUserFormProps {
...
-            key={form.key("email")}
-            {...form.getInputProps("email")}
-          />
-        </div>
+        <TextInput
+          withAsterisk
+          label="Email Address (Gmail only)"
+          placeholder="Example@gmail.com"
+          key={form.key("email")}
+          {...form.getInputProps("email")}
```
promatty, line 56: why do we only accept gmail again?

src/app/_components/admincomponents/admin-dashboard.tsx
```diff
@@ -3,81 +3,244 @@
...
     },
     validate: {
-      email: (value) => (value.trim().length > 0 ? null : "Email is required"),
-      organizationRole: (value) => (value.trim().length > 0 ? null : "Role is required"),
-      organizationId: (value) => (value.trim().length > 0 ? null : "Organization is required"),
+      email: (value) => {
+        if (value.trim().length === 0) return "Email is required";
+        const emailRegex = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
+        if (!emailRegex.test(value)) return "Invalid email address";
+        return null;
```
promatty, line 64: we have email regex in a shared function already. also, if we are only allowing gmail then the regex should also adjust for this

burtonjong:
@Yemyam I think once you address matthews comments and then change the "invite agency" to "create agency" this looks good

src/app/_components/admincomponents/invite-user-form.tsx
Yemyam: Oops, the email actually does not need to be gmail

src/app/_components/admincomponents/admin-dashboard.tsx
Yemyam: I'll change this to the shared function. I just checked the backend, and I don't think theres anything requiring the invitations to only work with gmail.

src/app/_components/admincomponents/invite-user-form.tsx
promatty: can we use colors from globals?

src/app/_components/admincomponents/admin-dashboard.tsx
```diff
@@ -3,81 +3,242 @@
...
+  };
 
-    inviteUserMutation.mutate(form.values);
+  const handleInviteTypeSelect = (type: "user" | "agency") => {
+    setInviteType(type);
   };
 
   return (
     <>
       <Button onClick={() => setShowInviteModal(true)}>Invite New User</Button>
```
promatty: can we make the button say "Send Invitation" instead of invite new user

src/app/_components/admincomponents/admin-dashboard.tsx
```diff
@@ -3,81 +3,242 @@
...
 
-  const form = useForm({
+  const createAgencyMutation = api.organization.createOrganization.useMutation({
+    onSuccess: (data) => {
+      if (!data) return;
+      notify.success(`Agency "${data.name}" created successfully`);
+      agencyForm.reset();
+      setShowInviteModal(false);
+      void organizations.refetch();
+    },
```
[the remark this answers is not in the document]
promatty: i mean i dont see this happening since we return the org object, but to be safe lets add error handling

promatty (approved):
small changes, approving now so you can merge after fixing them

PR #50, opened 2026-01-26
SANC-80: Driver Trip List API Endpoint

jason-duong4509:
Created an endpoint that returns all trips for a driver on a given date range. Also connected the endpoint to the front-end, allowing the information to display on the calendar

The video below shows the usage of the new endpoint when connected to the front-end. It also shows filtering by driver ID through assigning a trip to another driver in the DB

Each time a new date is viewed, the endpoint is queried for trips on that date

https://github.com/user-attachments/assets/808959f9-e50a-4b77-beec-adc26a35b37e

## Summary by CodeRabbit

* **New Features**
  * Driver dashboard added with an interactive calendar that loads trips for selected date ranges.
  * Calendar selection now updates displayed date range dynamically.

* **Bug Fixes / Improvements**
  * Loading indicator and calendar-hidden state while trip data loads.
  * Server-side date validation and driver-specific trip retrieval for accurate results.


src/app/_components/drivercomponents/driver-dashboard.tsx
```diff
@@ -0,0 +1,49 @@
...
+      startDate: dateJSON.startDate,
+      endDate: dateJSON.endDate,
+    },
+    { enabled: dateJSON.startDate !== "" && dateJSON.endDate !== "" },
+  );
+
+  if (tripQuery.data) {
+    //Query returned data from endpoint call
+    driverTrips = tripQuery.data as Booking[];
+  }
```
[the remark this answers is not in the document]
jason-duong4509: Leaving the initial `driverTrip` value as `[{}]` results in a blank calendar (a calendar with no trips). No errors appear during this period and a blank calendar is intentional behavior. Resolving the comment as a result

src/app/driver/home/page.tsx
burtonjong: we dont need this aynmore

src/app/driver/home/page.tsx
```diff
@@ -1,33 +1,12 @@
...
+import { DriverDashboard } from "@/app/_components/drivercomponents/driver-dashboard";
 import { requireRole } from "@/lib/auth-helpers";
 import { api, HydrateClient } from "@/trpc/server";
-import { type Booking, Role } from "@/types/types";
+import { Role } from "@/types/types";
 import styles from "./index.module.css";
 
 export default async function DriverHome() {
   const session = await requireRole([Role.ADMIN, Role.DRIVER]);
   const hello = await api.form.hello({ text: "from tRPC" });
```
burtonjong: we don't need this anymore either

src/server/api/routers/bookings.ts
```diff
@@ -250,4 +251,63 @@ export const bookingsRouter = createTRPCRouter({
...
+          message: "User not found in database",
+        });
+      }
+
+      if (user.role !== Role.DRIVER) {
+        throw new TRPCError({
+          code: "UNAUTHORIZED",
+          message: "User is not a driver",
+        });
+      }
```
burtonjong, line 303: what if on the admin page we want to be able to see bookings for a specific driver? this endpoint wouldn't work then, would it?

burtonjong (changes requested):

src/server/api/routers/bookings.ts
burtonjong: I guess we could just filter by driver id on the client side, but I think all endpoints should be accessible by admins regardless. Let me know what you think

promatty:
also resolve merge conflicts before we review please

src/app/driver/home/page.tsx
jason-duong4509: Sounds good. I removed it

src/app/driver/home/page.tsx
jason-duong4509: Gotcha. Removed this one too

src/server/api/routers/bookings.ts
jason-duong4509: As of right now, the endpoint is designed to return bookings that are assigned to the user who called it. If we extended this to admins, the endpoint would only return bookings assigned to the admin. I do not think this would be useful in this context

Achieving functionality where the admin could use the endpoint to filter bookings for a specific driver would require modification of the endpoint

src/server/api/routers/bookings.ts
```diff
@@ -275,4 +275,63 @@ export const bookingsRouter = createTRPCRouter({
...
+      const user = await ctx.db.query.user.findFirst({
+        where: (user, { eq }) => eq(user.id, ctx.session.user.id),
+      });
+
+      if (!user) {
+        throw new TRPCError({
+          code: "NOT_FOUND",
+          message: "User not found in database",
+        });
+      }
```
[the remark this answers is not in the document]
jason-duong4509: Keeping it the way it is because we should know whether the user's ID is in the DB before trying to return data pertaining to that user

src/server/api/routers/bookings.ts
```diff
@@ -275,4 +275,63 @@ export const bookingsRouter = createTRPCRouter({
...
+
+      if (user.role !== Role.DRIVER) {
+        throw new TRPCError({
+          code: "UNAUTHORIZED",
+          message: "User is not a driver",
+        });
+      }
```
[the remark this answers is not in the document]
jason-duong4509: done

burtonjong (approved):
k i think its fine then

PR #52, opened 2026-02-02
SANC-72: Add calendar event popover

themaxboucher:
https://codethechangeyyc.atlassian.net/browse/SANC-72

**Agency View**
Desktop:
Screenshot 2026-02-05 at 10 34 17 AM
Mobile:
Screenshot 2026-02-05 at 10 34 44 AM

**Admin View**
Desktop:
Screenshot 2026-02-05 at 10 27 49 AM
Mobile:
Screenshot 2026-02-05 at 10 28 22 AM

**Driver View**
Desktop:
Screenshot 2026-02-05 at 10 29 20 AM
Mobile:
Screenshot 2026-02-05 at 10 29 04 AM

## Summary by CodeRabbit

* **New Features**
  * Mobile drawer and desktop popover for viewing event details
  * Role-aware calendar modes (admin, driver, agency) with role-specific actions (edit/delete for admins; survey action for drivers)
  * Compact event tiles with time range and status indicator; expanded event detail view with addresses, passenger info, purpose

* **Style**
  * Rounded event corners; improved selection and focus visuals; adjusted past-date color

* **Chores**
  * Added an icon library and new status utilities (icons, colors, labels)


src/lib/utils.ts
```diff
@@ -0,0 +1,33 @@
+import { CircleCheck, CircleX } from "lucide-react";
+
+import { BookingStatus } from "@/types/types";
+
+export function getStatusIcon(status: BookingStatus | null) {
```
burtonjong: I think to be more descriptive it would be better to rename these functions to getBOOKING...

There could be other helper functions where we getSURVEYicons or similar so just add that keyword there

src/types/types.ts
```diff
@@ -49,6 +49,8 @@ export enum ViewMode {
   TABLE = "table",
 }
 
+export type CalendarViewType = "admin" | "driver" | "agency";
+
```
burtonjong, line 57: Think we should keep it consistent and either make viewmode a type, or make calendarviewtpye an enum. I'll leave that up to you, but im an enum liker

src/app/_components/common/calendar/event-details.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+        {pickupAddress && (
+          <Stack gap="0.25rem">
+            <Group gap="0.35rem" wrap="nowrap" c="dimmed">
+              <MapPin size={14} />
+              <Text size="xs">Pickup</Text>
+            </Group>
+            <Text size="sm">{pickupAddress}</Text>
+          </Stack>
+        )}
```
burtonjong: For these, I think instead of just rendering them if they exist, I think we should have some text to say whether or not they were even filled out

Imagine you were a user and one booking has that field and one doesn't, you might be wondering maybe if thats a system glitch or something, we should instead tell the user, for example if the pickupAddress is missing, say "{fieldName} was not filled out" or similar for the rest as well

src/app/_components/common/calendar/event-details.tsx
```diff
@@ -0,0 +1,139 @@
...
+interface EventDetailsProps {
+  title: string;
+  start: Date | null;
+  end: Date | null;
+  status: BookingStatus | null;
+  pickupAddress?: string;
+  destinationAddress?: string;
+  passengerInfo?: string;
+  purpose?: string | null;
+  viewType?: CalendarViewType;
```
burtonjong: I feel like this interface can be made into a type in types.ts

Like most of the things here are from the Booking type, so you can use Pick<Booking> & viewType and create a type out of that

This is only in consideration for when we eventually want to edit the EventDetails as well.

src/app/_components/common/calendar/event-block.tsx
```diff
@@ -0,0 +1,49 @@
...
+              </Flex>
+            </Tooltip>
+          )}
+        </Box>
+      </Group>
+    );
+  },
+);
+
+EventBlock.displayName = "EventBlock";
```
burtonjong: What is this?

src/app/_components/common/calendar/event-details.tsx
burtonjong: If this isn't possible or you think is uneccessary lmk

burtonjong (changes requested):

src/types/types.ts
themaxboucher: Made it an enum

src/app/_components/common/calendar/event-details.tsx
themaxboucher: I simplified things so that the `CalendarEvent` object is passed down directly instead of having everything copied to the props first.

src/app/_components/common/calendar/event-block.tsx
themaxboucher: It was there due to the `forwardRef` implementation of the component, but I simplified it so that the component simply accepts a `ref` prop. This is necessary so that the popover knows where to appear beside the calendar event block

src/app/_components/common/calendar/event-details.tsx
themaxboucher: Updated it so that it shows a hyphen if the info is missing. Let me know if this works.

Screenshot 2026-02-16 at 3 26 10 PM

themaxboucher:
I'm trying to figure out how to make the drawer closing nice. The Mantine component doesn't support swiping to close the modal. You have to click outside the drawer or click on an exit button in the drawer to close it. Neither are the ideal user experience. I'm thinking of using this drawer component instead: https://vaul.emilkowal.ski/. Thoughts?

With close button:
Screenshot 2026-02-16 at 4 13 49 PM

Without close button (not obvious how to close the drawer):
Screenshot 2026-02-16 at 4 21 21 PM

burtonjong (approved):

PR #54, opened 2026-02-04
SANC-85: Remove Homepage

jason-duong4509:
Removed the default homepage at `/` and changed the landing page from `/` to `/login`. Also updated a loading screen in the agency page to use the reusable loading element rather than a separate one

https://github.com/user-attachments/assets/71894bf4-6a1a-47d7-9ed7-0cce1cc67f6e

## Summary by CodeRabbit

* **New Features**
  * Users are now redirected from the site root ("/") to the login page.

* **Refactor**
  * Agency view now shows a dedicated loading screen instead of an inline placeholder.
  * Login flow timing updated: loading/redirect gating now uses local modal state to better coordinate session readiness.

* **Removed**
  * The default home page content has been removed and no longer renders.


promatty:
@jason-duong4509 merge conflict

burtonjong:
@promatty I had him do an extra bit thats why i didn't approve the first time

promatty (commented):
if i load the site while logged in, i still see the login modal before getting redirected. is there an easy fix to this?

jason-duong4509:
> if i load the site while logged in, i still see the login modal before getting redirected. is there an easy fix to this?

Fixed! Thanks for pointing this out. It previously worked until I broke it with the recent commit. The fix is simple but requires some thinking

The issue was that going to `/login` while logged in gives the same session behavior as going to `/login` after switching tabs. This causes both to display the modal only, while the former scenario should display the loading UI. With the most recent commit, every time session data changes, we always default to the last rendered UI (modal or loading) until we are 100% sure the user is/is not logged in; we change the UI accordingly when this happens

src/app/login/page.tsx
```diff
@@ -20,6 +20,18 @@ export default function LoginPage() {
   const router = useRouter();
   const { data: session, isPending: sessionLoading } = authClient.useSession();
   const hasRedirected = useRef(false);
+  const [hideLoginModal, setHideLoginModal] = useState(true);
```
burtonjong, line 23: Can't hide login modal just be
```suggestion
const shouldHideLogin = sessionLoading || !!session;
```

?

burtonjong (changes requested):

src/app/login/page.tsx
jason-duong4509: If I am understanding correctly, this will not fix the login error because `sessionLoading` is `true` during silent refreshes (when refetching authentication data to see if authentication status has changed). The original bug was caused by the code hiding the login modal during silent refreshes

burtonjong (approved):

PR #56, opened 2026-02-07
Sanc 28 implement driver trip assignment system

Lujarios:
## Summary by CodeRabbit

* **New Features**
  * Enhanced booking debug interface with day/driver selection, time inputs, and phone number field.
  * Added automatic end-time calculation for bookings.
  * Implemented driver availability checks with inline alerts and schedule visibility.
  * Added booking validation system that verifies driver availability and travel time between consecutive trips.

* **Refactor**
  * Optimized database performance with additional indexing.


.gitignore
```diff
@@ -9,7 +9,8 @@
 /coverage
 
 # database
-drizzle/meta 
+drizzle/meta
+drizzle
```
burtonjong: drizzle shouldnt be added

src/server/api/routers/bookings.ts
```diff
@@ -31,6 +324,7 @@ export const bookingsRouter = createTRPCRouter({
 
           // Optional fields
           purpose: z.string().optional(),
+          phoneNumber: z.string().max(25).optional().nullable(),
```
burtonjong, line 376: why are we adding phone number here?

src/server/api/routers/bookings.ts
```diff
@@ -2,18 +2,311 @@ import { TRPCError } from "@trpc/server";
...
+    endTime: string;
+    excludeBookingId?: number;
+  },
+): Promise<{ id: number } | undefined> {
+  const { driverId, startTime, endTime, excludeBookingId } = params;
+  const baseCondition = and(
+    eq(bookings.driverId, driverId),
+    ne(bookings.status, "cancelled"),
+    lt(bookings.startTime, endTime),
+    gt(bookings.endTime, startTime),
```
burtonjong, line 71: Eventually this query might be slow

It would be good to add an index on all 4 of these fields

burtonjong (changes requested):
!

src/server/api/routers/bookings.ts
Lujarios: Phone number was added to the booking schema outside this branch so I added it to the create route.

.gitignore
Lujarios: Fair. I had added it as I needed to delete some of the meta as it was pulling meta from other dev's pushes. I'll just keep this locally.

src/server/api/routers/bookings.ts
Lujarios: Good tip thanks!
image

.gitignore
Lujarios: Oh my bad meta was in there originally. I added drizzle as files like "drizzle\0002_daffy_valeria_richards.sql" would be the base for other "yarn generate" commands and when based off of other people's database progress it was messing with my personal database.

promatty:
merge conflicts @Lujarios

burtonjong (approved):

PR #66, opened 2026-03-08
SANC-91: Bookings SMS notifications

themaxboucher:
https://codethechangeyyc.atlassian.net/browse/SANC-91

Added SMS notifications using Telnyx whenever a booking is created or updated.

Example:
IMG_5097

## Summary by CodeRabbit

* **New Features**
  * Drivers receive SMS notifications when bookings are created and when key booking details (date/time, pickup, destination) change; update notifications only send when relevant fields differ.

* **Chores**
  * Added SMS delivery support via a Telnyx integration and introduced related environment variables and runtime dependency.


themaxboucher:
Currently, SMS notifications only get sent to `TEST_RECIPIENT_PHONE_NUMBER` defined in `.env`. Needs to be set up so that the notifications are sent to the phone numbers of all driver users.

src/server/api/routers/bookings.ts
```diff
@@ -11,6 +11,7 @@ import {
 } from "@/constants/driver-assignment";
 import { roundUpToNearestIncrement } from "@/lib/datetime";
 import { getTravelTimeMinutes } from "@/lib/google-maps";
+import { sendBookingUpdatedSms } from "@/lib/sms";
```
promatty, line 14: do we need to notify on bookings being created?

src/server/api/routers/bookings.ts
```diff
@@ -633,6 +634,28 @@ export const bookingsRouter = createTRPCRouter({
           .returning();
       });
```
promatty, line 646: maybe we should only notify on specific updates, like status (ie. if a ride is cancelled), start/end times changing, driverid change, or others you think are important. we just dont care to notify if there's a useless title change, etc

.env.example
```diff
@@ -25,4 +25,8 @@ NEXT_PUBLIC_APP_URL=
 
 # Google Maps
 NEXT_PUBLIC_GOOGLE_MAPS_API_KEY=
-GOOGLE_MAPS_API_KEY=
\ No newline at end of file
+GOOGLE_MAPS_API_KEY=
+
+# Telnyx (for SMS notifications)
+TELNYX_API_KEY=
+TELNYX_NUMBER=
```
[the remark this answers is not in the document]
promatty: ya just add this

src/lib/sms.ts
```diff
@@ -0,0 +1,60 @@
...
+  await Promise.all(
+    numbers.map(async (to) => {
+      try {
+        await conn.messages.send({
+          from: env.TELNYX_NUMBER,
+          to,
+          text: message,
+        });
+      } catch (err) {
+        console.error(`Failed to send SMS to ${to}:`, err);
```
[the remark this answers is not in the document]
promatty: ya lets redact since all users would be able to see phone num

src/server/api/routers/bookings.ts
```diff
@@ -633,6 +634,28 @@ export const bookingsRouter = createTRPCRouter({
...
+
+        await sendBookingUpdatedSms(
+          {
+            title: updated.title,
+            startTime: updated.startTime,
+            pickupAddress: updated.pickupAddress,
+            destinationAddress: updated.destinationAddress,
+          },
+          driverPhones,
+        );
```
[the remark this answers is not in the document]
promatty: yeah we should avoid this latency

src/server/api/routers/trip.ts
```diff
@@ -32,5 +35,23 @@ export const tripRouter = createTRPCRouter({
...
+
+      await sendBookingCreatedSms(
+        {
+          title: input.title,
+          startTime: input.startTime,
+          pickupAddress: input.pickupAddress,
+          destinationAddress: input.destinationAddress,
+        },
+        driverPhones,
+      );
```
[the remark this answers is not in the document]
promatty: avoid latency again

src/server/api/routers/bookings.ts
themaxboucher: I updated it so that SMS notifications only get sent out if the start time, pickup address, or destination address get updated. Let me know if this works or we want something different.

We could also just show the fields that have been updated in the text message. Something like:

> Updates have been made on booking #[booking id]:
> Start time: 12:00 PM -> 1:15 PM
> Pickup address: location 1 -> location 2

themaxboucher: Also now that there is the `driver_id` field in the `bookings` table. Would it make sense to only notify drivers about their own bookings? How do drivers get assigned to bookings in the first place?

src/server/api/routers/bookings.ts
themaxboucher: Yes, that code is in `trip.ts` since that is the mutation called when the bookings form is submitted. Is there a reason there are two different mutation to create a new booking?

src/server/api/routers/bookings.ts
promatty: yeah i like showing the fields that changed, if it isnt too complicated to do. and lets keep notifying all drivers; there are only ever 1 or 2 drivers doing the trips

src/server/api/routers/bookings.ts
promatty: oh i see, bookings.create is only for debugging. this is good then

themaxboucher:
Here is what the new SMS notifications look like:

IMG_5139

promatty (approved):
lgtm

PR #74, opened 2026-04-03
SANC-102: save user role to db

tanvimahal:
When inviting a new user, the user was being created in the user table but their role was never set, so they always defaulted to "driver". 
Added a db update after the user is created to explicitly set their role to whatever is selected (admin, driver, or agency).

## Summary by CodeRabbit

* **New Features**
  * User invitations to organizations now require specifying a role during the invitation process. When users are invited to join an organization, they will be assigned their designated role upon completing the invitation.


promatty (approved):
lgtm!

PR #78, opened 2026-04-14
SANC-108: Made agency name dynamic

jason-duong4509:
Added a new endpoint, getUserOrgName, to return a user's organization name from the DB. Used the new endpoint to return the user's organization name to the client, and had that info displayed in the nav bar

https://github.com/user-attachments/assets/cdfd2bee-361c-4c0f-83df-cdb2d03ee398

## Summary by CodeRabbit

* **Improvements**
  * Enhanced logout experience by ensuring cached user data is properly cleared, providing a clean state for subsequent login sessions.

### DIFF
--- a/src/app/_components/common/navbar.tsx
+++ b/src/app/_components/common/navbar.tsx
@@ -16,7 +16,6 @@ type NavbarView = "admin" | "agency" | "driver";
 
 interface NavbarProps {
   view: NavbarView;
-  agencyName?: string;
 }
 
 interface NavLinkProps {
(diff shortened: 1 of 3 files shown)


promatty (changes requested):
i found this bug:
1. log in as agency one
2. log out. log in as agency three
shows agency one as the name still 

Image

jason-duong4509:
> i found this bug:

Thanks for catching it! It turns out that the client's browser was caching data and reusing it briefly upon reentering the same page. Both the schedule data and organization name were stale and reused from a previous session

Going forward, upon logging out, the client's browser clears its cache so that reentering the same page as a different user will trigger a refetch of data. Before, this occurred only when the window was focused

